PR #119, opened 2018-02-26
Implement publisher service class.

smarlowucf:


mash/services/publisher/service.py
```diff
@@ -25,3 +37,422 @@ class PublisherService(BaseService):
...
+        if job.utctime != 'always' or job.status == SUCCESS:
+            self._publish_message(job)
+        job.listener_msg.ack()
+
+    def _publish_image(self, job_id):
+        """
+        Publish image for job that matches job_id
+        """
+        job = self.jobs[job_id]
+        job.publish_image(host=self.host)
```
rjschwei: host?

smarlowucf: Good catch. That is left over from old credentials model and can be cleaned up.

PR #120, opened 2018-02-27
Don't delete creds job if utctime is always.

smarlowucf:
I think we want utctime in credentials job doc right?

That way we can leave the job alone if 'always' job. Otherwise on last service the job is deleted. So as discussed an always job must be explicitly deleted using delete message from job creator.


schaefi:
:+1:

PR #125, opened 2018-02-27
Create key pair if it does not exist

smarlowucf:
If key pair does not exist create it in the region based on the provided private key.

I think this is much better then temp keys since we are given the information that is expected. Thus if it's not as expected we can fix it. No need to create temp keys on every test run.


schaefi:
I fixed the conflict and also enabled travis again for the repo as we are now public again :-)

The code looks good to me

smarlowucf:
> I fixed the conflict and also enabled travis again for the repo as we are now public again :-)

Thanks!

PR #126, opened 2018-02-27
No routing key use for uploader

schaefi:
Delete use of routing_key in uploader service
    
All information to classify a message is part of the message itself
Thus there is no reason to check on the routing key.

While testing the pipeline I stumbled over a couple of smaller issues for which I have created extra commits to this pull request which can be reviewed on a commit base

### DIFF
--- a/mash/services/uploader/service.py
+++ b/mash/services/uploader/service.py
@@ -116,64 +116,61 @@ class UploadImageService(BaseService):
                     )
                 }
             )
-        if message.method['routing_key'] == 'job_document':
+        if 'uploader_job' in job_data:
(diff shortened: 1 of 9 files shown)


mash/services/uploader/service.py
```diff
@@ -116,64 +116,61 @@ def _process_message(self, message):
                     )
                 }
             )
-        if message.method['routing_key'] == 'job_document':
+        if 'uploader_job' in job_data:
```
smarlowucf: We would still have routing_keys they would just be generic for credentials similar to the job docs. But this way of determining message type works just as well.

PR #129, opened 2018-02-28
Add replication service to spec.

smarlowucf:
- Update config files in spec. Only one config file for mash now.
- Fix missing testing service file install.


schaefi:
:+1:

PR #133, opened 2018-03-08
Move remove job config method to base class.

smarlowucf:


smarlowucf:
Agree, name isn't the best...

smarlowucf:
Looks like the dateutil plugin changed error format on latest update. Will fix in master and re-merge to open requests.

smarlowucf:
Okay, fixed now. Maybe we should pin the test requirements to minor versions to prevent this type of breakage when packages update changes that aren't compatible?

`python-dateutil~=2.7`

To get patch updates but prevent breaking updates.

schaefi:
This changes looks good to me :+1:

PR #135, opened 2018-03-08
Remove ssh key name requirement from testing.

smarlowucf:
IPA [will add](https://github.com/SUSE/ipa/pull/68) public key to EC2 instance based on private key file using cloud init script.

Private key file will be provided as a configuration item in MASH config file.

### DIFF
--- a/config/mash_config.yaml
+++ b/config/mash_config.yaml
@@ -13,3 +13,5 @@ non_cred_services:
   - obs
 obs:
   download_directory: /var/tmp/mash/images
+testing:
+  ssh_private_key_file: /etc/mash/testing_key
--- a/mash/services/testing/config.py
+++ b/mash/services/testing/config.py
@@ -16,6 +16,7 @@
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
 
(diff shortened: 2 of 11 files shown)


schaefi:
As discussed by mail this looks good to me. There is just a conflict caused by the merge of the simplification patch for the testing service you provided too. Feel free to merge once conflict got resolved

config/mash_config.yaml
```diff
@@ -16,3 +16,5 @@ credentials:
   credentials_directory: /var/lib/mash/credentials/
 obs:
   download_directory: /var/tmp/mash/images
+testing:
+  ssh_private_key_file: /etc/mash/testing_key
```
rjschwei: We do not want to write the private key to disk if it can be avoided and for mash, since it is generated, based on the discussions this week, it should live in /var/lib/mash/keys/ssh

smarlowucf: The idea is that this is a setting and not something that needs to be auto generated. In setting up a MASH pipeline we setup a ssh private/public key pair and point to the file location. `/var/lib/mash/keys/ssh/testing_key` could be default but it's nice to give the user options.

This can be discussed in #164 .

PR #136, opened 2018-03-09
Rename source_regions in replication job doc.

smarlowucf:
- Rename to replication_source_regions to be consistent with other msgs.
- Remove unused image_id and ssh_user args from replication jobs.
- Update credential handling per cred service changes.
- Cleanup validation args in replication service.

### DIFF
--- a/examples/messages/replication_job.json
+++ b/examples/messages/replication_job.json
@@ -4,7 +4,7 @@
         "image_description": "My Image",
         "provider": "ec2",
         "utctime": "now",
-        "source_regions": {
+        "replication_regions": {
             "us-east-1": {
                 "account": "test-aws",
                 "target_regions": ["us-east-2", "us-west-2", "eu-west-3"]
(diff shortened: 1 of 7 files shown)


rjschwei:
Hm, I think the term "source_region" is more precise than "replication_region" a "replication_region" can be interpreted as a "to" or "from" region as in "replicate to replication_region X" or "replicate from replication_region Y" however it would be strange to state "replicate to source_region Z" when "source" reasonably clearly expresses that it should be the origin.

smarlowucf:
source_regions is also confusing because we are passing the same key with different data type in all the listener messages.

E.g. [testing result](https://github.com/SUSE/mash/blob/master/examples/messages/testing_result.json). Which is passed into replication service.

rjschwei:
Just because we use the same name in different places does not necessarily mean we have to have the same structural representation of what that name points to. While such macro level consistency might be nice it is not a necessity nor is it necessarily easy to come by.

Anyway, if you must, then name it replication_sources or replication_source_regions

smarlowucf:
I like `replication_source_regions`, that is the most explicit.

smarlowucf:
> Anyway, if you must, then name it replication_sources or replication_source_regions

Made the change. Going ahead with merge.

PR #139, opened 2018-03-12
Add publisher service to spec.

smarlowucf:
- Update project source repo link.
- Match python version in setup.py and tox config.


schaefi:
:+1:

PR #140, opened 2018-03-13
Fix bug in publish job result method.

smarlowucf:
- Messages should be published to listener queue instead of service queue.


schaefi:
:+1:

PR #141, opened 2018-03-13
Use base remove file method.

smarlowucf:
Update publisher and credentials services to use base remove file method.


schaefi:
:+1:

PR #144, opened 2018-03-13
Do not log exception outside try/except.

smarlowucf:
- Log error when stack trace not available.
- Update provider name to be lower case in add job methods.


schaefi:
:+1:

PR #145, opened 2018-03-13
Use encryption_keys_file instead of private key file.

smarlowucf:
- With symmetric key encryption no need for public and private keys.
- File will contain a list of encryption keys (initially just one) that will eventually be used in key rotation.

### DIFF
--- a/mash/services/base_defaults.py
+++ b/mash/services/base_defaults.py
@@ -27,6 +27,10 @@ class Defaults(object):
         return '/etc/mash/mash_config.yaml'
 
     @classmethod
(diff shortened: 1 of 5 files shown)


mash/services/base_defaults.py
```diff
@@ -26,6 +26,10 @@ class Defaults(object):
     def get_config(self):
         return '/etc/mash/mash_config.yaml'
 
+    @classmethod
+    def get_encryption_keys_file(self):
+        return '/etc/mash/encryption_keys'
```
rjschwei: Fix location

PR #146, opened 2018-03-14
Add missing docstrings and order methods

smarlowucf:
Commits are split so they can be reviewed separately.

- Add missing docstrings.
- Reorder methods in alpha order.

### DIFF
--- a/mash/services/base_service.py
+++ b/mash/services/base_service.py
@@ -130,8 +130,14 @@ class BaseService(object):
         self._publish(exchange, job_id, message)
 
     def consume_queue(self, callback, queue_name=None):
+        """
+        Declare and consume queue.
+
+        If queue_name not provided use service_queue name attr.
+        """
         if not queue_name:
             queue_name = self.service_queue
+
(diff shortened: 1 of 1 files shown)


schaefi:
Well, looks like inspired by Robert :) I never understood why the order of methods matters. The name has to match what the method does and the names should be meaningful, but the order ? who cares for that ?

Anyway I know the discussion will be pointless so I'll just merge this and hope no context change is involved because any order change like that is a mess to review which is also one reason why I would not do that.

Just one thing, please let us do this once and then never again. It's just pain for no benefit and questionable reasons

Thanks

smarlowucf:
>  inspired by Robert

No, just like things in order if possible. There's a lot of methods in base so it helps scanning code and being able to locate things when they are in some sort of order.

> Anyway I know the discussion will be pointless so I'll just merge this and hope no context change is involved because any order change like that is a mess to review which is also one reason why I would not do that.

Simply added missing docstrings then re-ordered methods. No functional changes made.

PR #147, opened 2018-03-14
Add credentials service methods for backend.

smarlowucf:
Credentials encrypted with encryption key and stored in separate files.


schaefi:
So cryptography.fernet is our choice as it seems. Looks good for me but I think we should give Robert a chance to review too

smarlowucf:
> So cryptography.fernet is our choice as it seems. Looks good for me but I think we should give Robert a chance to review too

Yeah, good to have agreement here. I'm no expert in choosing encryption methods. Cryptography seems to be the best python crypto library and Fernet is their recommended option for encrypting strings.

smarlowucf:
Note: Depending on the decision here the file path generation should be moved to a separate method to prevent duplication and making changing the path structure easier.

```
path = os.path.join(
    self.credentials_directory, user, provider, account
)
```

smarlowucf:
Updated per discussion. We will use the MultiFernet types and save keys in a file. This will enable the possibility for key rotation in the future.

smarlowucf:
Opening a new request with base and credentials service methods .

PR #150, opened 2018-03-15
Handle exceptions in base decode method.

smarlowucf:
- Exceptions are logged and None returned if credentials invalid.
- Process and return job_id + credentials instead of dictionary message from decode method.


rjschwei:
LGTM, lets have @schaefi give this a once over as well.

PR #156, opened 2018-03-26
Remove source region from results after replication

smarlowucf:
No need to pass source regions after replication. Publishing, deprecation and PINT service use image name.

### DIFF
--- a/examples/messages/deprecation_result.json
+++ b/examples/messages/deprecation_result.json
@@ -2,12 +2,6 @@
     "deprecation_result": {
         "id": "123",
         "cloud_image_name": "image_123",
-        "source_regions": {
-            "us-east-2": "ami-bc5b48d0",
-            "us-west-2": "ami-bc5b48d0",
-            "eu-west-2": "ami-bc5b48d0",
-            "cn-northwest-1": "ami-bc5b4853"
-        },
         "status": "success"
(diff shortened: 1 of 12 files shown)


rjschwei:
For replication we need a "source region" which I think we decided should be called "replication_source_regions" how does this relate to the pull request where the name was changed?

smarlowucf:
This is from the discussion we had about listener messages. I was sending the source regions to publisher -> deprecation. However, we agreed that those three services plus pint use the cloud image names not id's. So title is confusing I suppose:

"Remove source_region info from listener messages after the testing service."

So there's no need to pass source_regions in replication listener message. And so on for every service downstream in pipeline.

E.g. the listener messages (results messages) can be simplified from:

```
{
    "replication_result": {
        "id": "123",
        "cloud_image_name": "image_123",
        "source_regions": {
            "us-east-2": "ami-bc5b48d0",
            "us-west-2": "ami-bc5b48d0",
            "eu-west-2": "ami-bc5b48d0",
            "cn-northwest-1": "ami-bc5b4853"
        },
        "status": "success"
    }
}
```

To:

```
{
    "replication_result": {
        "id": "123",
        "cloud_image_name": "image_123",
        "status": "success"
    }
}
```

PR #160, opened 2018-03-26
Return consistent collections type.

smarlowucf:
In base config get services list. Don't use sets to preserve order of services when returning non credential services.

### DIFF
--- a/test/unit/services/base/config_test.py
+++ b/test/unit/services/base/config_test.py
@@ -27,7 +27,7 @@ class TestBaseConfig(object):
         assert self.config.get_private_key_file() == '/etc/mash/creds_key'
 
     def test_get_services_names(self):
-        # Non credential services
+        # Services requiring credentials
         expected = [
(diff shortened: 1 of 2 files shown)


test/unit/services/base/config_test.py
```diff
@@ -27,11 +27,17 @@ def test_get_private_key_file(self):
         assert self.config.get_private_key_file() == '/etc/mash/creds_key'
 
     def test_get_services_names(self):
-        expected = {
-            'publisher', 'pint', 'testing', 'replication',
-            'uploader', 'deprecation'
-        }
+        # Non credential services
```
rjschwei: This comment is misleading. It could be interpreted as "All services but  the credential service" or "all services that do not need credentials" (although this list appears to be all services that need credentials). The comment should probably be "# Services requiring credentials"

schaefi:
This is imho good to be merged as Sean took care for the change requested. Will do the merge and hope it's ok

PR #162, opened 2018-03-28
Make keywords json valid to be used in testing.

smarlowucf:
Rename accounts to groups and relation to accounts. Accounts map one-to-one with a partition. So mold the data for how it will be used.

### DIFF
--- a/examples/keywords.json
+++ b/examples/keywords.json
@@ -1,26 +1,18 @@
-# Provides the connection between keyword processing for the job and
-# expansion to known entities per framework.
-# PLACEHOLDERS are in all CAPS
 {
-  'ec2': {
-    'regions': {
-      'all': ['aws','aws-cn','aws-us-gov'],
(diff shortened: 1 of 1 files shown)


examples/accounts.json
```diff
@@ -0,0 +1,38 @@
...
+    "groups": {
+      "GROUP_ONE": ["ACCOUNT_ONE", "ACCOUNT_TWO"]
+    },
+    "accounts": {
+      "ACCOUNT_ONE": "aws-cn",
+      "ACCOUNT_TWO": "aws"
+    },
+    "helper_images": {
+      "ap-northeast-1": "ami-383c1956",
+      "ap-northeast-2": "ami-249b554a",
```
rjschwei: ap-northeast-3 => ami-82444aff

smarlowucf:
@rjschwei gtg?

PR #164, opened 2018-03-29
Use temp ssh keys in testing service.

smarlowucf:
Key is created based on the supplied private key file prior to testing in the testing account. After testing the key is removed.

### DIFF
--- a/config/mash_config.yaml
+++ b/config/mash_config.yaml
@@ -17,4 +17,4 @@ credentials:
 obs:
   download_directory: /var/tmp/mash/images
 testing:
-  ssh_private_key_file: /etc/mash/testing_key
+  ssh_private_key_file: /etc/mash/key
--- a/mash/services/testing/ec2_job.py
+++ b/mash/services/testing/ec2_job.py
@@ -32,7 +32,7 @@ class EC2TestingJob(TestingJob):
     def __init__(
         self, id, provider, ssh_private_key_file, test_regions, tests, utctime,
         job_file=None, credentials=None, description=None, distro=None,
-        instance_type=None, ssh_user='ec2-user'
+        instance_type='t2.micro', ssh_user='ec2-user'
     ):
         super(EC2TestingJob, self).__init__(
             id, provider, ssh_private_key_file, test_regions, tests, utctime,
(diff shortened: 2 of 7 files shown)


mash/services/testing/ec2_job.py
```diff
@@ -32,7 +32,7 @@ class EC2TestingJob(TestingJob):
     def __init__(
         self, id, provider, ssh_private_key_file, test_regions, tests, utctime,
         job_file=None, credentials=None, description=None, distro=None,
-        instance_type=None, ssh_user='ec2-user'
+        instance_type='t2.micro', ssh_user='ec2-user'
```
rjschwei: The instance type default should remain 'None'.

If a user wants to test on a specific type they have to specify it. Otherwise the test will be subject to the instance type randomizer i.e. we want:

```
if not instance_type:
    self.instance_type = choose_random_test_instance()
```

I don't think we want to delegate the randomization to the client.

config/mash_config.yaml
```diff
@@ -17,4 +17,4 @@ credentials:
 obs:
   download_directory: /var/tmp/mash/images
 testing:
-  ssh_private_key_file: /etc/mash/testing_key
+  ssh_private_key_file: /etc/mash/key
```
rjschwei: We don't want to store the private key. If we have to store the private key to make paramiko happy then it should go into /var/lib/mash/keys/.... The uploader and tester both need ssh keys and the default config should store the keys in the same location. Since the keys are generated and temporary they should be considered run time application data and thus should end up in /var/lib/mash

mash/utils/generic.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_key_from_file(key_file_path):
+    """
+    Return a key as string from the given file.
+    """
+    with open(key_file_path, 'r') as key_file:
+        key = key_file.read().strip()
+
+    return key
```
rjschwei: filename, generic?  -> utils, mash_utils

smarlowucf: It's a generic file with generic utility methods :smile: .

mash/services/testing/ec2_job.py
smarlowucf: Will revert the instance type changes here for now. The random instance types functionality is a separate todo.

smarlowucf:
As a note: The random instance type task is marked as MASH next so it is on the radar.

smarlowucf:
Things to determine:

- [ ] Are we sticking with temp EC2 key-pairs?
- [ ] If so then we are okay using a singe private/public key?
  - [ ] Then ssh_private_key_file can move from a testing config value to a MASH project level config value for use by uploader and testing services.

config/mash_config.yaml
smarlowucf: @rjschwei Moved changes to a separate PR. #187

smarlowucf:
@rjschwei Good?

PR #172, opened 2018-04-03
Remove validation from replication job classes.

smarlowucf:


rjschwei:
OK for conflict resolution then merge.

PR #173, opened 2018-04-03
Remove validation from publisher job classes.

smarlowucf:


rjschwei:
Resolve merge conflict then OK to merge

PR #175, opened 2018-04-03
Remove notify invalid config handling (testing).

smarlowucf:
In testing service class.


rjschwei:
Expecting separate PR for remove of validation at service level.

PR #176, opened 2018-04-04
Remove notify invalid config handling (replication).

smarlowucf:
From replication service.


rjschwei:
Merging, expect validation removal in another PR

PR #177, opened 2018-04-04
Remove notify invalid config handling (publisher).

smarlowucf:
From publisher service.


rjschwei:
Merging expect validation removal in separate PR

PR #184, opened 2018-04-04
Attempt creating log dir if it does not exist.

### LINKED ISSUE #183: Setup log directory
When starting a service the log file is initialized with a new log handler. If the directory does not exist this just raises an exception.

It would be nice to check for the directory and create it if it doesn't exist.

https://github.com/SUSE/mash/blob/02d59d50f5ac9dc84e86c6b3e7c904bd77bc7519/mash/services/base_service.py#L359

smarlowucf:
When setting log handler. File handler can create the file if it does not exist but not the directory.

Resolves #183

### DIFF
--- a/mash/services/base_service.py
+++ b/mash/services/base_service.py
@@ -361,6 +361,10 @@ class BaseService(object):
         Allow to set a custom service log file
         """
         try:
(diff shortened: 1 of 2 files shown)


schaefi:
:+1:

PR #186, opened 2018-04-05
Remove service job doc validation

smarlowucf:
From:

- Testing
- Credentials
- Replication
- Publisher


schaefi:
I would wait with this until the schema is merged and actively playing

PR #190, opened 2018-04-09
Use temp ssh keys for uploader in ec2

schaefi:
On upload temporary create and register an ssh keypair

### DIFF
--- a/mash/services/uploader/cloud/amazon.py
+++ b/mash/services/uploader/cloud/amazon.py
@@ -15,11 +15,15 @@
 # You should have received a copy of the GNU General Public License
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
+from tempfile import NamedTemporaryFile
+from collections import namedtuple
 from ec2utils.ec2uploadimg import EC2ImageUploader
 
 # project
 from mash.services.uploader.cloud.base import UploadBase
 from mash.mash_exceptions import MashUploadException
+from mash.utils.ec2 import get_client
+from mash.utils.mash_utils import generate_name
 
 
 class UploadAmazon(UploadBase):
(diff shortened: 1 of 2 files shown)


rjschwei:
This works, bt based on implementation and comment there is still a misunderstanding I think.

The "temporary part" was primarily centered around what shows up in EC2. To maybe help eliminate the confusion/misunderstanding let me try and out line this as steps:

1.) Upon startup if no ssh key exists in /var/lib/mash/ssh_key mash creates a public/private ssh key pair
2.) The uploader creates a temporary random key-pair name
3.) Using the public key from step 1 and the name form step 2 the uploader uses ec2.import_key_pair()
4.) delete key-pair in EC2 when done

calling ec2.create_keypair() and then downloading the private key is another option to solve the problem.

I'm have no problem with this implementation but thought it was important to point out the other possible approach.

schaefi:
There is no misunderstanding. I already understood that the decision was made that mash creates a ssh key pair and stores that such that this pubkey can be send to ec2. To be honest I wanted to free myself from external references and I have a problem with the level of trust.

In the last meeting I suggested mash creates a ssh keypair and imports the pubkey under a generic name to ec2. This would have been simple but yes assumes nobody plays bad with the imported keys on the ec2 side. Actually it seems I was the only one who wouldn't have a problem with that level of trust as others in the call had concerns or made fun of me.

So that's ok but now mash creates a ssh keypair and stores it locally and imports the same pubkey under a random name several times which puts permanent trust on the locally stored private key. I'm sorry but if there is no trust in the key pool on the ec2 side then there is also no trust on the private key on the mash server side.

I think we cheat ourselves if we import the same key just with another name and consider that a temporary key. Either it's really temporary or it should be simple.

Just to explain why I implemented it this way

smarlowucf:
> Actually it seems I was the only one who wouldn't have a problem with that level of trust as others in the call had concerns or made fun of me.

I don't think that was the case. I think all the options we discussed aren't ideal and have disadvantages. The best option using cloud init script we decided we had to wait on. If we use a generic name I think it is more error prone since the user could delete the key at any time.

Anyway, I think both options that we chose work just as well. So this can be merged and eventually we will be able to clean it all up and use a simple cloud init script which would better match the GCE and Azure usage.

schaefi:
>eventually we will be able to clean it all up and use a simple cloud init script which would better
>match the GCE and Azure usage.

you have my vote on this

PR #191, opened 2018-04-10
Add encryption to credentials responses

smarlowucf:
- Add encryption and decryption methods for credentials using Fernet.
- Update existing services that require credentials to set the encryption keys file used for encryption/decryption.


smarlowucf:
Added the credentials service functionality to handle add account requests.

PR #192, opened 2018-04-12
Add json schema for add account messages.

smarlowucf:
Add account message that comes from API to job creator. To add an account for the requesting user.


mash/services/jobcreator/schema.py
```diff
@@ -17,6 +17,34 @@
...
+            'properties': {
+                'access_key_id': {'$ref': '#/definitions/non_empty_string'},
+                'secret_access_key': {'$ref': '#/definitions/non_empty_string'}
+            },
+            'additionalProperties': False
+        },
+        'group': {'$ref': '#/definitions/non_empty_string'},
+        'partition': {'$ref': '#/definitions/non_empty_string'},
+        'provider': {'enum': ['azure', 'ec2']},
+        'requesting_user': {'$ref': '#/definitions/non_empty_string'},
```
rjschwei: This is not going to work. I think we should create "add_account_ec2" and "add_account_azure" to account for the different properties. In the current implementation "azure" and "ec2" are allowed for "add_account" but "access_key_id" and "secret_access_key" may not be empty, this makes no sense for azure when ewe have a publish settings file, or other means of authentication.

smarlowucf: Okay will rename this for EC2.

smarlowucf: So these would be different message schemas specific to provider. We probably want to handle them with different keys and drop the provider field.

Messages would arrive by key like add_account_ec2/add_account_azure?

smarlowucf: Actually, I prefer a single key in messages add_account and an explicit provider value. So we could use `if message['provider'] == 'ec2': validate with ec2 doc` etc.

PR #193, opened 2018-04-12
Implement job creator.

smarlowucf:
- Handles job initiation message. Generates and publishes a message for all services if job is valid.
- Handles add account message. Adds account to accounts file and sends a message to credentials service with encrypted credentials if account valid.

Note: There's a bit of overlap here with existing PR's so that the branch passes build and has the required methods. The main concern here is the job_creator service itself.


smarlowucf:
TODO: Handle delete job message in service message method...

schaefi:
I think jobcreator/service.py is too big. Every job doc is in its own method but I think there should be a bit more structure and design. Actually I think each job doc deserves to be a class. It would give contributors the opportunity to do conflict free changes as well as the readability will be improved. Also if needed methods can be implemented per job document and have a clear scope.

The code is good but I think there is room for improvement on style and readability

schaefi:
@rjschwei please give this one a look to and let us decide if we go forward or not. Thanks

smarlowucf:
> The code is good but I think there is room for improvement on style and readability

Agreed. Looking a little into the Azure integration I realize this isn't very extensible the way it is. Going with separate classes as you suggested would, I think better encapsulate the handling for each job doc per service.

Should this be refactored or noted to do so with Azure integration?

smarlowucf:
> I personally would refactor now before merge

I think that's the better option. Get it right now rather than merging code with refactor TODO's. However, this will require us being able to and knowing what's needed for Azure publishing. Which right now is a blocker. So this would be on hold until that's cleared up.

smarlowucf:
Going ahead and closing this. Will work on refactoring the job doc handling into a cleaner class structure. And break down into more bite size chunks.

PR #199, opened 2018-04-16
Rename ipa_helper to ec2_ipa_helper.

smarlowucf:
- Pass ssh_user to base job class.
- Move temp key pair handling to a context manager.

GCE and Azure will have similar credential handling (ipa_helper) and eventually EC2 will follow suit.


smarlowucf:
> For this difference I have a hard time see why we would want to separate, what should amount to a simple if condition into different files just to push the if condition which we need regardless to some other place?

In going that route it becomes multiple if conditions. Using regular methods you have to check the provider twice on setup and in finally after test.

So I thought the separate modules was cleaner and easier to read. But it leads to a lot of duplication.

PR #200, opened 2018-04-17
Add azure job class.

smarlowucf:
- Azure and GCE will use service account files for authentication. Use ipa_helper module name for both.

### DIFF
--- a/mash/services/testing/ipa_helper.py
+++ b/mash/services/testing/ipa_helper.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2018 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# mash is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with mash.  If not, see <http://www.gnu.org/licenses/>
+#
+
+import logging
+import threading
+
+from ipa.ipa_controller import test_image
+
+from mash.services.status_levels import EXCEPTION, FAILED, SUCCESS
+from mash.utils.mash_utils import temp_file
+
+
+def ipa_test(
+    results, provider=None, access_key_id=None, description=None, distro=None,
+    image_id=None, instance_type=None, region=None, secret_access_key=None,
+    service_account_credentials=None, ssh_private_key_file=None, ssh_user=None,
+    tests=None
+):
+    name = threading.current_thread().getName()
+    status = 1
+    try:
+        with temp_file(service_account_credentials) as service_account_file:
+            status, result = test_image(
+                provider,
+                cleanup=True,
+                desc=description,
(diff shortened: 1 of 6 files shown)


test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
...
+
+        self.testing._add_job({'id': '1', 'provider': 'azure'})
+
+        mock_create_job.assert_called_once_with(
+            AzureTestingJob, {'id': '1', 'provider': 'azure'}
+        )
+
     def test_testing_add_job_exists(self):
         job = Mock()
         job.id = '1'
```
rjschwei: here job.id gets assigned

test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
...
         job.get_metadata.return_value = {'job_id': '1'}
 
         self.testing.jobs['1'] = Mock()
-        self.testing._add_job({'id': '1', 'provider': 'ec2'})
+        self.testing._add_job({'id': '1', 'provider': 'azure'})
```
rjschwei: and here the same literal is used, I think the value should be 'job.id' rather than using the literal

test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
             EC2TestingJob, {'id': '1', 'provider': 'ec2'}
         )
 
+    @patch.object(TestingService, '_create_job')
+    def test_testing_add_azure_job(self, mock_create_job):
+        job = Mock()
+        job.id = '1'
```
rjschwei: assignment for literal

test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
...
+        job.get_metadata.return_value = {'job_id': '1', 'provider': 'azure'}
```
rjschwei: using the same literal instead of job.id

test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
...
+
+        self.testing._add_job({'id': '1', 'provider': 'azure'})
```
rjschwei: dito

test/unit/services/testing/service_test.py
```diff
@@ -84,13 +85,25 @@ def test_testing_add_job(self, mock_create_job):
...
+
+        mock_create_job.assert_called_once_with(
+            AzureTestingJob, {'id': '1', 'provider': 'azure'}
```
rjschwei: and once again use of the literal for the id

mash/services/testing/ipa_helper.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) 2018 SUSE Linux GmbH.  All rights reserved.
```
rjschwei: I would expectthe introduction of ipa_helper to trigger a change in ec2_job to take advantage of the new utility?

smarlowucf: Sorry this is confusing. I created two helper modules to handle the two different ipa flows. Temp key and temp service accounts files.

I will just make this and #200 be a single request to use the same helper with if clauses rather than separate modules and context managers.

test/unit/services/testing/service_test.py
smarlowucf: Will clean these up.

PR #206, opened 2018-04-20
Consume listener messages in uploader service.

smarlowucf:


smarlowucf:
Sorry @schaefi, I broke this when I changed the base publish result method.

schaefi:
yes looks like, thanks for fixing it... would have taken me longer to find that ;)

PR #207, opened 2018-04-20
Integrate encrypted credentials with response.

smarlowucf:
Credentials response will be a dictionary mapping accounts to encrypted credential strings.


smarlowucf:
> I just trust that _get_encrypted_credentials() does what we expect :)

It just reads the encrypted string from file based on the result of _get_credentials_file_path and returns the string as unicode (default) with stripped whitespace.

PR #213, opened 2018-04-23
Add pretty print log of new job docs.

smarlowucf:


smarlowucf:
> If so please open an issue, but I guess it has not a high prio at the moment

I think it'd be nice to have at least similar output. But this is low prio IMHO. We should take some time during the workshop to do some code review and refactor.

rjschwei:
The initial idea was that we'd dump the complete job request json to the log file as the header. That should give us all the information we need in the log file. With this change we are going to replicate json information in the log file I think. Not sure we really want to do this.

rjschwei:
| So this would happen in job creator when a job has been validated?

Yes

smarlowucf:
Noted

PR #218, opened 2018-05-03
Implementation of upload service for Azure

### LINKED ISSUE #197: Azure support for uploader
So far the uploader backend supports EC2 but should also support Azure

* https://docs.microsoft.com/en-us/python/azure/python-sdk-azure-authenticate?view=azure-python#mgmt-auth-file

schaefi:
Add UploadAzure class and integration to uploader service
This Fixes #197


schaefi:
During testing I used an uploader job description as follows:

```json
{
    "uploader_job": {
        "id": "123",
        "cloud_image_name": "ms_image",
        "image_description": "marcus image",
        "utctime": "now",
        "provider": "azure",
        "target_regions": {
            "westeurope": {
                "resource_group": "ms_group",
                "container_name": "ms1container",
                "storage_account": "ms1storage"
            }
        }
    }
}
```

From a credentials point of view I "duck typed" the credentials service in a way that it contains the dictionary to match the Azure auth file requirements. This means for completion of the Azure support I think the job creator and also the credentials service needs azure related patches

smarlowucf:
> and also the credentials service needs azure related patches

The credentials service really should require no changes. If job creator encrypts the credentials when an account is added then the credentials service will just store them based on account, user, provider.

schaefi:
>The credentials service really should require no changes. If job creator encrypts the credentials when >an account is added then the credentials service will just store them 

ok, so it's the job creator who needs to know the details what makes up Azure credentials. That's fine for me and I guess you will handle that part, right ?

schaefi:
> We'll want to update the example messages

yes forgot that, thanks for pointing it out

schaefi:
>...Is this something that is entered for each job or for each new account?

I'd say that is job specific. With one Azure account (in my case I used our azure-rd account) it's possible to reach all regions of the Azure platform the account has been created for.

> Is it something that is region >specific in Azure or across regions?

So resource_group -> storage_account -> storage_container are all connected to the region. In Azure it's not possible to put something into a storage container which is not in the storage_account of the resource_group in that region. That's why I think our uploader_job layout has been designed quite well

> And is it something that could be expected to be the same name if it's different across regions?

I did not understand that question

smarlowucf:
> That's fine for me and I guess you will handle that part, right ?

Yep I will work on this. It is related to whether the config items are account or job based. See #219 and #220. These would be account specific. I think maybe this would be easier to have another MASH call to discuss the options here?

mash/services/uploader/cloud/__init__.py
```diff
@@ -60,6 +61,11 @@ def __new__(
                 credentials, system_image_file, cloud_image_name,
                 cloud_image_description, custom_uploader_args
             )
+        if csp_name == CSP.azure:
```
eandersson: Minor, but this should probably be an elif.

schaefi: yes, thanks :+1:

schaefi:
>Anyway, we could provide the option for both account based and per job if we want. And maybe I'm >overlooking something? Maybe we should setup a short call to go over Azure a bit (I know Azure >publishing is on hold for now in Azure ARM)?

Azure publishing is not clear to me too as there were several changes between the arm and asm model and also the new and old gallery concept. So yes regarding this topic we should have a call at best with James Mason involved.

> I guess a better question is, given Azure ARM won't require replication would you always want
>to use the same region and bucket when releasing new images for an account? If there's no >replication then there's not the same problem of choosing a random target/source region and thus it >could be something that's specific to an account?

I think the uploader interface should not depend on how later publishing will be performed and should be flexible to upload anywhere whereas the target endpoint is described by the account info from credentials and the upload job description. In Azure you can reach any endpoint from a storage point of view by the correct Azure account credentials, the region, the default or given resource group, the storage account name and the container name. I think the uploader for Azure should be able to deal with all that information when given. So yes I think we should stay flexible here and allow to specify a target_region for Azure in the job document too. This also makes it clear where we upload to.

smarlowucf:
@schaefi In your example:

```
{
    "uploader_job": {
        "id": "123",
        "cloud_image_name": "ms_image",
        "image_description": "marcus image",
        "utctime": "now",
        "provider": "azure",
        "target_regions": {
            "westeurope": {
                "resource_group": "ms_group",
                "container_name": "ms1container",
                "storage_account": "ms1storage"
            }
        }
    }
}
```

I think we will need the account as part of each target region. So we can connect the account name with credentials response.

schaefi:
don't get that, I would expect all account related information from the credentials response

schaefi:
Ah now I understand, yes I overlooked that part. Thanks and fixed

smarlowucf:
> I think the uploader interface should not depend on how later publishing will be performed and should be flexible to upload anywhere whereas the target endpoint is described by the account info from credentials and the upload job description.

Well it's partly dependent. That is the publishing on the uploader. The jobcreator for Azure will provide the target region which is based on the job or account.

The uploader will upload to that region and testing + publisher rely on the image being in the proper region.

schaefi:
I consider this one done and ready for merge

Thoughts ?

mash/services/uploader/conventions/__init__.py
```diff
@@ -34,6 +35,8 @@ class Conventions(object):
     def __new__(self, csp_name):
         if csp_name == CSP.ec2:
             return ConventionsAmazon()
+        if csp_name == CSP.azure:
```
eandersson: Minor but this should be elif as well

schaefi: yes agreed, will fix

mash/services/uploader/cloud/azure.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) 2017 SUSE Linux GmbH.  All rights reserved.
```
rjschwei: 2018

mash/services/uploader/cloud/azure.py
```diff
@@ -0,0 +1,106 @@
...
+        if not self.custom_args:
+            self.custom_args = {}
+
+        if 'region' in self.custom_args:
+            self.region = self.custom_args['region']
+        else:
+            raise MashUploadException(
+                'required Azure region name for upload not specified'
+            )
+
```
rjschwei: For AWS we assign a default region, the Azure uploader should not have different behavior, i.e. we can add a default region here as well or we can change the AWS implementation to also throw an exception if no region is given. I am fine either way, but I think the behavior should be consistent across frameworks in this case.

mash/services/uploader/cloud/azure.py
```diff
@@ -0,0 +1,106 @@
...
+        if 'container_name' in self.custom_args:
+            self.container_name = self.custom_args['container_name']
+        else:
+            raise MashUploadException(
+                'required Azure container name for upload not specified'
+            )
```
rjschwei: Suggest:

```
if not self.custom_args.get('container_name'):
    raise ...

self.container_name = self.custom_args['container_name']
```
Saves the else clause. Similar for other occurrences.

mash/services/uploader/cloud/azure.py
```diff
@@ -0,0 +1,106 @@
...
+        storage_key_list = storage_client.storage_accounts.list_keys(
+            self.resource_group, self.storage_account
+        )
+        page_blob_service = PageBlobService(
+            account_name=self.storage_account,
+            account_key=storage_key_list.keys[0].value
+        )
+        page_blob_service.create_blob_from_path(
+            self.container_name, self.cloud_image_name, self.system_image_file
+        )
```
rjschwei: upload should end here, what follows is image creation.

The create_blob_from_path should have a "max-connections" or similar argument as on the command line. We'll want to experiment with the number, but the default is 1 which means an image upload to azure could take 4 or 5 hours. Example from my network at home:
- without using max-connections on the commandline upload takes about 8 hours
- with max-connections set to 16 upload takes about 8 hours
- with max-connections set to 4 upload takes about 30 minutes

So there is a lot to gain by finding the "right" value depending on the origination network. WHich also makes me think this should go into a configuration file.

mash/services/uploader/conventions/azure.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2017 SUSE Linux GmbH.  All rights reserved.
```
rjschwei: 2018

rjschwei:
On 05/04/2018 04:50 AM, Marcus Schäfer wrote:
>> ...Is this something that is entered for each job or for each new account?
> 
> I'd say that is job specific. With one Azure account (in my case I used our azure-rd account) it's possible to reach all regions of the Azure platform the account has been created for.
> 
>> Is it something that is region >specific in Azure or across regions?
> 
> So resource_group -> storage_account -> storage_container are all connected to the region. In Azure it's not possible to put something into a storage container which is not in the storage_account of the resource_group in that region. That's why I think our uploader_job layout has been designed quite well
> 
>> And is it something that could be expected to be the same name if it's different across regions?
> 
> I did not understand that question
> 

No, we should not expect those resources to have the same name across
regions. While it would make sense for a user to have lets say
"image_storage" for example as a storage account in any region the user
may upload images to, this can certainly not be expected.

mash/services/uploader/cloud/azure.py
schaefi: Will not change that because you run into a stack trace if you access a not existing key from a dictionary in python. As custom_args can contain any information the object constructing code validates them and I'd like to have the opportunity to provide a better error message than the dict raise

mash/services/uploader/cloud/azure.py
schaefi: > The create_blob_from_path should have a "max-connections" ...

In the sdk from multiapi/storage/v2017_07_29/blob/blockblobservice.py max connections is set to 2. I guess there might be different default values depending on which version of the storage sdk is in use. So I'm ok if we set one. If this should be in a configuration file is something we should address in an extra issue. For now I'll set max_connections=4

schaefi: > upload should end here, what follows is image creation.

Why should we stop after blob upload ? After upload the next service is testing. Testing needs to start an instance of an image. If upload does not report something an instance can be started from, testing and publishing would have no reference. Sounds weird to me, and inconsistent regarding how amazon upload works (reporting an ami-id to start an instance from)

mash/services/uploader/cloud/azure.py
rjschwei: That's the point of the get() method over "direct" access, try it ;)

-> python3
Python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = {}
>>> a.get('p')
>>>

rjschwei:
On 05/30/2018 04:55 AM, Marcus Schäfer wrote:
> schaefi commented on this pull request.
> 
> 
> 
>> +        self._create_auth_file()
> +
> +    def upload(self):
> +        storage_client = get_client_from_auth_file(
> +            StorageManagementClient, auth_path=self.auth_file.name
> +        )
> +        storage_key_list = storage_client.storage_accounts.list_keys(
> +            self.resource_group, self.storage_account
> +        )
> +        page_blob_service = PageBlobService(
> +            account_name=self.storage_account,
> +            account_key=storage_key_list.keys[0].value
> +        )
> +        page_blob_service.create_blob_from_path(
> +            self.container_name, self.cloud_image_name, self.system_image_file
> +        )
> 
>> The create_blob_from_path should have a "max-connections" ...
> 
> In the sdk from multiapi/storage/v2017_07_29/blob/blockblobservice.py max connections is set to 2. I guess there 
> might be different default values depending on which version of the storage sdk is in use. So I'm ok if we set one. 
> If this should be in a configuration file is something we should address in an extra issue. For now I'll set 
> max_connections=4
> 


Fair enough, filed #232

Thanks

rjschwei:
On 05/30/2018 05:11 AM, Marcus Schäfer wrote:
> schaefi commented on this pull request.
> 
> 
> 
>> +        self._create_auth_file()
> +
> +    def upload(self):
> +        storage_client = get_client_from_auth_file(
> +            StorageManagementClient, auth_path=self.auth_file.name
> +        )
> +        storage_key_list = storage_client.storage_accounts.list_keys(
> +            self.resource_group, self.storage_account
> +        )
> +        page_blob_service = PageBlobService(
> +            account_name=self.storage_account,
> +            account_key=storage_key_list.keys[0].value
> +        )
> +        page_blob_service.create_blob_from_path(
> +            self.container_name, self.cloud_image_name, self.system_image_file
> +        )
> 
>> upload should end here, what follows is image creation.
> 
> Why should we stop after blob upload ? After upload the next service is testing. Testing needs to start an instance 
> of an image. If upload does not report something an instance can be started from, testing and publishing would have 
> no reference. Sounds weird to me, and inconsistent regarding how amazon upload works (reporting an ami-id to start 
> an instance from)
> 

Sorry my comment was to terse, yes we need an image. But we have a
method that is called "upload" but the end effect of that method is that
an "image" is created. So either end the "upload" method after uploading
and create a "create_image" method and call "upload" form withing the
new method or rename the existing method to "create_image"

Having an "upload" method that creates an image smels too much like
"side effect programming"

schaefi:
Sorry I disagree. The upload method exists in the scope of the uploader service and specifies a global term to upload an image as we build it into a cloud. This upload defines a process including all steps such that an instance can be created. The upload method is a method from the UploadAzure
class and the class docs tells its responsibility

You named your Amazon tool implementing such a process: "ec2uploadimg" and it does a lot more than what the name indicates, it even spawns instances. I'm sorry if you change your naming I'll change mine.

rjschwei:
On 05/30/2018 09:41 AM, Marcus Schäfer wrote:
> Sorry I disagree. The upload method exists in the scope of the uploader service and specifies a global term to upload an image as we build it into a cloud. This upload defines a process including all steps such that an instance can be created. The upload method is a method from the UploadAzure
> class and the class docs tells its responsibility
> 
> You named your Amazon tool implementing such a process: "ec2uploadimg" and it does a lot more than what the name 
> indicates, it even spawns instances. I'm sorry if you change your naming I'll change mine.

There is clearly a difference between the name of a tool (ec2uploadimg)
and the name for a method inside an implementation. If you look at the
ec2uploadimg implementation you will find methods such as
_create_snapshot, _attach_volume, create_image etc. and each of these
methods does as it is named without side-effects.

This method named "upload" actually creates and image.

But whatever....

schaefi:
The service is called uploader, we all agreed on that. What uploading means is different per cloud provider. Thus I designed the interface uploader/cloud/base.py to provide an upload() method which implements the process to upload. The technical details are not relevant at this point if the idea is to define a name for a process which is common across all clouds.

From a user perspective it's pretty clear that one must upload something. That this could include the creation of an image at the cloud provider side is a technical detail but not relevant in the scope of the service and its uploader interface.

Our uploader service creates a result. This result always includes an identifier needed to create an instance from the uploaded image. If you argue on me because of upload returning an image id this should consequently have an impact on the name of the service in general.

If you argue uploader service creating cloud images it should be renamed to what ? image service ? which then puts us back on arguing why an image service uploads. If we strictly set a name policy such that the last step in a process has to define the name you are all right, then everything needs to be renamed to create_image.

But may I question what's the point ?

* we agreed on uploader as the name for the service
* we have the uploader/cloud/base.py interface in place since ec2 upload was implemented
* we have uploader/cloud/amazon:upload() returning the ami id and agreed on it
* we have specified examples/messages/uploader_result.json to contain an image id

why is upload as name for the process returning an id now evil ?

Do you think it increases the motivation to contribute to a project when decisions made some time ago are not respected or end with a "whatever" message. I never question the names, the people who did the implementation chose, unless they are completely misleading, which I think is not the case here.

rjschwei:
On 05/30/2018 10:57 AM, Marcus Schäfer wrote:
> The service is called uploader, we all agreed on that. What uploading means is different per cloud provider. Thus I 
> designed the interface uploader/cloud/base.py to provide an upload() method which implements the process to upload. 
> The technical details are not relevant at this point if the idea is to define a name for a process which is common 
> across all clouds.
> 
> From a user perspective it's pretty clear that one must upload something. That this could include the creation of an 
> image at the cloud provider side is a technical detail but not relevant in the scope of the service and its uploader 
> interface.

Well it kind of is relevant. As a reader of the code, and being
knowledgeable about that fact that it takes multiple steps to get to the
point where one has a usable image it is surprising that the "upload"
method actually creates an image. When there are at least 2 steps in
every framework, upload something and form that create a usable image.

In common language "upload" is from my point of view similar to "copy".
I have something in one place and I will put it in another. Generall if
I were to us "cp a b" on my system as as a side effect I would get a.bck
on my system I'd be surprised. In the same way that I am surprised to
see that "upload" does a "create".

And there is nothing wrong with having, at higher levels, an "upload"
method, although in hindsight that should have probably been
"upload_image". However in the detailed implementation one would expect
the steps that it takes in a given cloud to get to a usable image to be
broken up into methods named accordingly  where each method does one thing.

> 
> Our uploader service creates a result. This result always includes an identifier needed to create an instance from 
> the uploaded image. If you argue on me because of upload returning an image id this should consequently have an 
> impact on the name of the service in general.

No, because as a user the idea of "uploading an image" is that there is
an image at the end of the process. So conceptually at that level
everything fits together. This does not necessitate that at the
implementation level, the closer we get to what we actually do, we carry
the same concepts. At the implementation level we need to use concepts
that make sense and invoke appropriate thought models for developers,
those that might read the code.

> 
> If you argue uploader service creating cloud images it should be renamed to what ? image service ? which then puts 
> us back on arguing why an image service uploads. If we strictly set a name policy such that the last step in a 
> process has to define the name you are all right, then everything needs to be renamed to create_image.

No I am arguing that at the appropriate level of abstraction we use the
appropriate names for what we do. Let me try with an analogy.

When a person that knows nothing about a car needs to get their car
towed to the shop to have it fixed they will say "my car is broken". At
that level of abstraction that's OK, to the layman the car is either
drive-able as a whole or it is not. Thus when I cannot drive it, it is
broken, as a whole. However to the mechanic the car is not broken, to
the mechanic it may be a loose wire on the starter, or a dead battery,
or whatever. And when the customer picks up the car the repair ticket
does not read "fixed car", but it will read "replaced battery".

What I am trying to express as that we want to use the appropriate names
and abstractions at the appropriate level. The service is the image
uploader, and I expect a functioning image at the end, but on the inside
of that beast there are a number of steps we have to complete, one of
those steps is uploading and another is creating and image.

> 
> But may I question what's the point ?
> 
> * we agreed on uploader as the name for the service

Different level of abstraction and I agree at that level the name makes
perfect sense and can easily be communicated to people less familiar or
completely unfamiliar with what it takes to create an image.

> * we have the uploader/cloud/base.py interface in place since ec2 upload was implemented
> * we have uploader/cloud/amazon:upload() returning the ami id and agreed on it
> * we have specified examples/messages/uploader_result.json to contain an image id
> 
> why is upload as name for the process returning an id now evil ?
> 
> Do you think it increases the motivation to contribute to a project when decisions made some time ago are not 
> respected or end with a "whatever" message. I never question the names, the people who did the implementation chose, 
> unless they are completely misleading, which I think is not the case here.

It is surprising.

And often time we do not stumble across everything the first time.

Anyway, obviously me being surprised by the side effect is obviously
causing more harm then good.

PR #220, opened 2018-05-03
Add Azure account examples.

smarlowucf:
@rjschwei Given Azure has no replication in ARM I don't think we need the list of regions per "partition". So the idea is that when an Azure account is added to MASH we expect the; target region, resource group, container name and storage account.

The upload would always go to the same location based on that info for always/persistent jobs.

### DIFF
--- a/examples/accounts.json
+++ b/examples/accounts.json
@@ -39,6 +39,24 @@
       "us-west-2": "ami-f0091d91"
     }
   },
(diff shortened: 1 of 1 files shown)


schaefi:
@rjschwei This hooks into the Azure account discussion by mail and we probably want to have a short meeting

PR #222, opened 2018-05-15
Implement job creator message handling

smarlowucf:
Split into subsequent commits so they can be reviewed separately:

- Rename job schema for EC2 only, each provider will have different requirements and options.
- Add base and EC2 job classes for handling service job docs
- Add factory method to handle validation and creation of job class instances
- Update service class to integrate service message handling of new jobs
- Update/add unit tests


smarlowucf:
Making a few changes regarding account/group handling in jobcreator...

PR #224, opened 2018-05-16
Create accounts file if it does not exist

smarlowucf:
- Add accounts template
- Create file from template if file does not exist

### DIFF
--- a/mash/services/jobcreator/service.py
+++ b/mash/services/jobcreator/service.py
@@ -17,9 +17,11 @@
 #
 
 import json
+import os
 
 from mash.services.base_service import BaseService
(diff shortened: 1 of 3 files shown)


mash/services/jobcreator/service.py
```diff
@@ -58,6 +64,15 @@ def _handle_service_message(self, message):
 
         message.ack()
 
+    def _write_accounts_to_file(self, accounts):
+        """
+        Update accounts file with provided accounts dictionary.
+        """
+        account_info = json.dumps(accounts, indent=2)
```
rjschwei: Be considered to old folks indent=4

PR #228, opened 2018-05-18
Integrate add account message handling.

smarlowucf:
Account messages will be sent to listener queue:

- The account is added to the accounts file.
- The credentials are encrypted.
- And the message is passed to credentials services.


schaefi:
Here I did not start a review as I think account management we agreed to be done by the credentials service. So I think the responsibility of the jobcreator service is to just send along a request to the credentials service instead of fiddling with account data itself

smarlowucf:
Yep, this will move. Closing for now.

PR #233, opened 2018-06-04
Integrate encryption key rotation

smarlowucf:
- The keys will be rotated buy a cron job running in background
scheduler.
- Rotation will occur every other week at 0000 on Sat.
- If there is an error with any credentials files the old keys
will remain in keys file.
- Once a successful rotation occurs all old keys are purged
leaving one encryption key.


smarlowucf:
@schaefi Yeah, it is not needed for initial release.

PR #238, opened 2018-06-05
Move static EC2 info to the yaml config.

### LINKED ISSUE #235: EC2 Region Info
This info should move to the mash config file instead of the dynamic json file.

Includes partitions and helper images.

smarlowucf:
Step 1: Move the ec2 region info to yaml config. 

- I think this is okay staying in the job creator since it's not really account specific.
- The job classes will handle this data the same way.
- Also made the "provider" generic in case there is a need for GCE or Azure info. This way it can be handled generically in factory method.

Fixes #235

### DIFF
--- a/config/mash_config.yaml
+++ b/config/mash_config.yaml
@@ -19,3 +19,48 @@ jobcreator:
   accounts_file: /etc/mash/accounts.json
 obs:
   download_directory: /var/tmp/mash/images
+provider:
+  ec2:
+    regions:
+      aws:
+        - ap-northeast-1
+        - ap-northeast-2
+        - ap-northeast-3
+        - ap-south-1
+        - ap-southeast-1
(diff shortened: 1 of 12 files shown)


config/mash_config.yaml
```diff
@@ -19,3 +19,48 @@ jobcreator:
   accounts_file: /etc/mash/accounts.json
 obs:
   download_directory: /var/tmp/mash/images
+provider:
+  ec2:
+    regions:
+      aws:
+        - ap-northeast-1
+        - ap-northeast-2
+        - ap-northeast-3
```
rjschwei: Please make a note somewhere or open an issue that ap-northeast-3 needs to move to "additional regions" implementation

smarlowucf: Noted

PR #240, opened 2018-06-07
Decompress the Azure image for upload.

smarlowucf:
@schaefi @rjschwei This still needs update to unit tests but does this look okay? Using shutil to do a buffered decompress and copy?

All Azure images will need this done so I think this fits in the Azure upload class.

### DIFF
--- a/mash/services/uploader/cloud/azure.py
+++ b/mash/services/uploader/cloud/azure.py
@@ -15,6 +15,10 @@
 # You should have received a copy of the GNU General Public License
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
+
+import lzma
+import shutil
+
 from tempfile import NamedTemporaryFile
 
 from azure.common.client_factory import get_client_from_auth_file
@@ -69,8 +73,14 @@ class UploadAzure(UploadBase):
             account_name=self.storage_account,
             account_key=storage_key_list.keys[0].value
         )
+
+        decompressed_name = self.system_image_file.rsplit('.', maxsplit=1)[0]
+        with lzma.LZMAFile(self.system_image_file) as compressed:
+            with open(decompressed_name, 'wb') as decompressed:
+                shutil.copyfileobj(compressed, decompressed)
+
         page_blob_service.create_blob_from_path(
-            self.container_name, self.cloud_image_name, self.system_image_file,
+            self.container_name, self.cloud_image_name, decompressed_name,
             max_connections=4
         )
         compute_client = get_client_from_auth_file(


schaefi:
Hmm, no I wouldn't do that. If we change the azure uploader to support compressed streams we should do it right from the beginning on. This means the following steps:

1. simply copy my ./utils/xz.py from the azurectl sources to mash
2. instead of page_blob_service.create_blob_from_path() use page_blob_service.create_blob_from_stream() and put the opening of the file into mash code

This will then automatically detect the file type open the stream appropriately, does chunk based decompression and the performance is ok

I can do that once we have sorted out all other blocking azure issues. Feel free top open an issue assigned to me, I'll jump on it as soon as I have a free slot

smarlowucf:
The LZMAFile used as context manager provides a buffer stream which uses LZMADecompressor which decompresses data incrementally. This appears to be what `./utils/xz.py` does right? Why not use builtins?

> This will then automatically detect the file type open the stream appropriately

This doesn't hurt but will we ever get an image from OBS that isn't compressed and that we don't need to decompress to upload to Azure?

What about something like:

```
if self.system_image_file.rsplit('.', maxsplit=1)[1] == 'xz':
    f_type = lzma.LZMAFile
else:
    f_type = open

with f_type(self.system_image_file) as f:
    page_blob_service.create_blob_from_stream(
        self.container_name, self.cloud_image_name, f,
        max_connections=4
    )
```

rjschwei:
We might have to support Stemcell uploads created by the CF pipeline and I am not certain we will get a compressed file out of that. So a detection is not necessarily a waste of time. Another expansion on the horizon would be to feed mash with a locally built image. However both of these cocepts apply universally and I do not necessarily want to implement or design for these at this point, YAGNI.

I am fine with using LZMA

schaefi:
> I am fine with using LZMA

we are using LZMA. The question is if the implementation I did with 
lzma.LZMADecompressor can be completely replaced with lzma.LZMAFile

reading the docs I don't see where lzma.LZMAFile implements an incremental decompression
which allows me to specify the buffer size. Maybe I just overlooked something.

Of course if simple opening with lzma.LZMAFile does the reading in chunks as we
need it I would use this.

Please also take into account that the code I have used here has already proven to
work in the context of azure uploading during the time we used azurectl. I think just
making use of it will be the most easy approach.

While investigating all this I was also reading the azure sdk code and I see they have
implemented the zero page handling but I could not see that they have implemented the
retry of pages as we have it in azurectl. The constant timeout issues I have when uploading
go away with the retry page code we have in azurectl. This might be only because of my
bad connection but just to mention we had a lot in azurectl which made our upload life easier

smarlowucf:
> reading the docs I don't see where lzma.LZMAFile implements an incremental decompression
> which allows me to specify the buffer size. Maybe I just overlooked something.

It's briefly mentioned for LZMADecompressor which the LZMAFile uses for reading, "which can be used to decompress data incrementally". But looking at the impl in cpython it ends up using a DecompressReader which is a buffer.

In testing this was the case.

> While investigating all this I was also reading the azure sdk code and I see they have
> implemented the zero page handling but I could not see that they have implemented the
> retry of pages as we have it in azurectl. The constant timeout issues I have when uploading
> go away with the retry page code we have in azurectl. This might be only because of my
> bad connection but just to mention we had a lot in azurectl which made our upload life easier

Not sure if it can handle retry on timeouts.

It's probably worth giving the different options a test since this is going to be the biggest bottleneck in the pipeline.

smarlowucf:
create_blob_from_stream for pageBlob requires a count parameter which is the size of buffer to read. Given the size is unknown in our case I think we will need two steps. First decompress to local disk then create blob from file or stream.

schaefi:
Ah now that you say it I remember why I wrote the page blob iterator using the update_page() method from the sdk directly. That also implements the mentioned page retry. See here:

* https://github.com/SUSE/azurectl/blob/master/azurectl/storage/page_blob.py

Are we open to re-use this code ? I think we should.

rjschwei:
Yes, we should reuse the code from azurectl

smarlowucf:
> Are we open to re-use this code ? I think we should.

Actually if Azure forces the images to be 30G and 512 aligned then all decompressed files would be the same size? Or is the 512 alignment the only hard rule?

Given the size of the files maybe it is worth having better fault tolerance with retry on error/timeouts which you have in Azurectl.

Ideally we would:

- Create blob directly from LZMA decompression stream and not require temp storage on disk.
- Don't send empty pages, which is covered with either route.
- Have ability to retry update of pages if timeout or errors occur.

So we could use the LZMAFile along with page_blob impl from azurectl to accomplish this?

rjschwei:
The alignment is a requirement, the size is not. For testing purposes we could easily set up smaller images, the 30GB is mandated for release.

smarlowucf:
> The alignment is a requirement, the size is not.

Okay. Good to use code from azurectl and LZMAFile if we can.

schaefi:
I have updated #241 accordingly

smarlowucf:
Closing: superseded by #241

PR #241, opened 2018-06-08
Added XZ stream handler and FileType class

schaefi:
In preparation to auto detect compressed images and
support uploading of decompressed chunks the XZ
and FileType classes were added


schaefi:
Here we go, that classes had been copied from my azurectl tool. I knew it would be helpful at some point :-)

schaefi:
References #240

smarlowucf:
See comment in #240, I think we should use the builtin LZMAFile type. Since it appears to provide the same functionality.

Having a class to handle filetype and basename is nice though.

rjschwei:
If LZMA does what we need it to do, we should use it, I agree.

schaefi:
I've added the support based on the "old" azurectl approach. Don't consider this as final as we still have a discussion on lzma usage in #240

mash/services/uploader/cloud/azure.py
```diff
@@ -69,9 +71,9 @@ def upload(self):
             account_name=self.storage_account,
             account_key=storage_key_list.keys[0].value
         )
-        page_blob_service.create_blob_from_path(
-            self.container_name, self.cloud_image_name, self.system_image_file,
-            max_connections=4
+        page_blob_service.create_blob_from_stream(
```
smarlowucf: @schaefi As a note from testing this method I don't think it will work for us. The method requires a count parameter which would be the size of file.

Since we buffering a decompressed file incrementally we don't know what the size will be.

smarlowucf: I think we'd need to copy the file to local disk decompressed unless there's another option for pulling straight from incremental buffer.

mash/services/uploader/cloud/azure.py
```diff
@@ -71,10 +74,15 @@ def upload(self):
...
-        )
+        if system_image_file_type.is_xz():
+            open_image = lzma.LZMAFile
+        else:
+            open_image = open
+        with open_image(self.system_image_file, 'rb') as image_stream:
+            page_blob_service.create_blob_from_stream(
+                self.container_name, self.cloud_image_name,
+                image_stream, max_connections=4
+            )
```
smarlowucf: The one method that the XZ module provides that's not in LZMAFile is the `uncompressed_size`. I think we could just use a static method based on stream like:

```
with open_image(self.system_image_file, 'rb') as image_stream:
    file_fize = stream_size(image_stream)
    page_blob_service.create_blob_from_stream(
        self.container_name, self.cloud_image_name,
        image_stream, max_connections=4, count=file_size
    )

...

def stream_size(stream):
    stream.seek(0, os.SEEK_END)
    size = stream.tell()
    stream.seek(0)
    return size
```

This works in testing, however if we want to guarantee no buffer issues the stream could be opened twice. Once to seek for size then to create blob.

mash/mash_exceptions.py
```diff
@@ -144,3 +144,27 @@ class MashValidationException(MashException):
...
+
+class MashAzurePageBlobSetupError(MashException):
+    """
+    Exception raised if the page blob can not be created
+    """
+
+
+class MashAzurePageBlobUpdateError(MashException):
+    """
+    Exception raised of the page blob can not be updated
```
rjschwei: "if" not "of" typo

mash/services/uploader/cloud/azure_page_blob.py
```diff
@@ -0,0 +1,126 @@
...
+
+        data = data_stream.read(requested_bytes)
+
+        if not data:
+            raise StopIteration()
+
+        length = len(data)
+        page_end = self.page_start + length - 1
+
+        if not data == zero_page:
```
rjschwei: Data comparison "!+" operator should work, is easier to read than "not ... ==..."

mash/services/uploader/cloud/azure_page_blob.py
```diff
@@ -0,0 +1,126 @@
...
+                        self.blob_name,
+                        data,
+                        self.page_start,
+                        page_end
+                    )
+                    break
+                except Exception as e:
+                    upload_errors.append(
+                        '{0}: {1}'.format(type(e).__name__, e)
+                    )
```
rjschwei: Wouldn't we get the same error message for uploading on failure?

The point being that the error messages are going to end up in a log somewhere and if they are all the same we should not print the message 5 times, rather we should collect the error message once and keep track of the repeat count separately. Could also take advantage of the retry_decorator.

schaefi: > Wouldn't we get the same error message for uploading on failure?

I don't know it which is the reason why it's just collected. The message comes from blob_service.update_page(). That is a method of the azure sdk and that method gets the information from the azure API. I don't know what information is exposed by the API. You are probably right and it will be the same message 5 times but I don't know exactly. I'll come up with a change that sort out duplicates and also adds an information how often we tried

schaefi: done now

PR #242, opened 2018-06-12
Move account handling to credential service.

### LINKED ISSUE #236: Account handling in credentials service
Account handling should move to credentials service from job creator.

JC will ping credentials service with job to ensure requesting user has the requested accounts. Credentials service will provide non-secret info regarding partition, special regions etc.

smarlowucf:
- Job creator validates job doc then request account check for job.
- Credentials service confirms the requesting user has the proper accounts.
- If accounts are valid job creator will create instance of class and send job docs to required services.

Still a few steps to do but this just moves existing workflow to credentials service. Also, I think the uuid for the jobs should be generated by the CLI and web interface when they are integrated. This way a user will have an ID from each request for reference. For now it's just generated by job creator service.

Resolves #236


rjschwei:
I think this moves us in the right direction. I had another look at accounts.json and I think the group definition is not yet correct as "requesting_user" should be a list. For example access to resale-images should be granted to a group of people on our team

smarlowucf:
> I think the group definition is not yet correct as "requesting_user" should be a list.

Well the decision was that the credentials were stored based on requesting user. So both rjschwei and smarlow would upload the resale-images credentials and they would need a unique name:

rjschwei/ec2/rs_resale_images
smarlow/ec2/sm_resale_images

And the groups would just be "psuedo groups" mapping to multiple accounts.

If the groups and accounts are being shared this changes the storage model we agreed on.

PR #247, opened 2018-07-10
Handle add account messages.

smarlowucf:
- Process add account request messages in credentials service.
- Accounts are added to json file and the credentials are encrypted and stored.
- Update unit tests for account handling changes.

### DIFF
--- a/mash/services/credentials/base_account.py
+++ b/mash/services/credentials/base_account.py
@@ -50,7 +50,7 @@ class BaseAccount(object):
         if groups:
             group = groups.get(self.group_name)
 
-            if group is None:
+            if not group:
                 groups[self.group_name] = [self.account_name]
             elif self.account_name not in group:
                 # Allow for account updates, don't append multiple times.
--- a/mash/services/credentials/base_account.py
+++ b/mash/services/credentials/base_account.py
@@ -1,0 +1,61 @@
+# Copyright (c) 2018 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


mash/services/credentials/base_account.py
```diff
@@ -0,0 +1,61 @@
...
+        Add the account to the group for the requesting user.
+
+        If the group does not exist create it with the new account.
+        """
+        groups = accounts[self.provider]['groups'].get(self.requesting_user)
+
+        if groups:
+            group = groups.get(self.group_name)
+
+            if group is None:
```
schaefi: Could be changed to

```python
if not group:
```

schaefi: I think empty strings are also considered a no group which are not matched when looking for None type

test/unit/services/credentials/base_account_test.py
```diff
@@ -0,0 +1,29 @@
...
+    def setup(self):
+        self.account = BaseAccount(
+            'acnt123', 'aws', 'ec2', 'user3', group_name='group123'
+        )
+
+    def test_add_account(self):
+        self.account.add_account({'test': 'data'})
+
+    def test_add_multiple_accounts(self):
+        with open('../data/accounts.json') as f:
```
schaefi: nitpick but ```f``` is a bad name :smile:

mash/services/credentials/base_account.py
smarlowucf: Yeah, this would only return a dict or None but that works.

smarlowucf:
@schaefi Made the suggested changes, going ahead with merge.

PR #249, opened 2018-07-12
OBS no api access

### LINKED ISSUE #217: Support osc release feature in obs service
The obs service downloads the image build from a given project. This requires an account in the build service used for the mash use case and also a role setup for that account in the projects handled by the obs service

A better solution to the account problem could be to use the osc release feature. See the following excerpt from a conversation by mail which summarizes the idea

```
OK, from my point of view we have our answer, publish into a separate
project and get the binaries refreshed with "osc release". According to
the mail thread with Adrian that should afford us a credential less
access to the binaries and the project does not need a new generic role
associated with it.

Of course this requires a bunch of changes in the obs service and it
requires some preliminary work as well. I have a hard time believing the
"osc release" can be run without credentials. Thus this should be tested
first before any code changes are made.

If that works the obs service would instead of locking the package, run
"osc release" and then pick up the package from the alternate release area.

The upside is that we do not have to handle credentials for OBS at all
and any user that wants to use mash can configure their OBS project with
a couple of lines of metadata.
```

schaefi:
__Refactor obs service using download server not api__

As discussed in the workshop:

The obs service was based on using the api via the osc commandline tool. However that requires osc credentials to be setup which can't be done without opening a wide gate to access and manipulate the project(s) in obs.  From the mash perspective all we need is access to the binary build results and its metadata. Both exists on  the published download area. Thus this patch refactors  the obs service to target the download server contents instead of the accessing the api through osc.

This Fixes #217


schaefi:
Please note the obs job description has a different meaning now. There was an optional key api_url which has changed to still optional download_root. Also the meaning of the image key has changed
as the following example shows:

```json
{
    "obs_job": {
        "id": "123",
        "download_root": "http://download.suse.de/ibs",
        "image": "SLES12-Azure-BYOS",
        "project": "Devel:/PubCloud:/Stable:/Images12",
        "utctime": "now",
        "conditions": [
            {"package": ["kernel-default", ">=4.13.1", ">=1.1"]},
            {"image": "1.42.1"}
        ]
    }
}

* image: specifies the name as we set it in the kiwi XML description
* project: specifies an uri path such that it is a valid location, see the :/ notation
* download_root: defaults to the public obs download area, must be set if the project lives somewhere else
```

schaefi:
I'm pretty sure the change has an impact on the job creator service. Thus this should not be forgotten

schaefi:
Also please note we lost all of our locking concept as it would require api access which we don't want anymore. As discussed I tried to add a "fail safe" use of data in a way that the first request reads the metadata file (.packages). If all constraints are valid the actual image and it's shasum file gets downloaded. This is the same as before, but after download a check runs which compares the version and build number from the very first metadata download. If there was any change the service treats the download as faulty. When obs publishes new results either version of build number changes. So the concept is based on dealing with data which matches those.

smarlowucf:
> I'm pretty sure the change has an impact on the job creator service. Thus this should not be forgotten

Yep, will add an issue to update jobcreator and the message examples to reflect these changes when they are merged.

mash/services/obs/build_result.py
```diff
@@ -185,29 +190,24 @@ def get_image(self):
...
-                        os.sep.join([self.download_directory, binary])
-                    )
-                except Exception as e:
-                    raise MashImageDownloadException(
-                        'Image Download failed with: {0}'.format(e)
-                    )
-        return downloaded
+        build_number = self._get_build_number(self.image_metadata_name)
+        image_files = self.remote.fetch_files(
+            self.image_name, ['.xz', 'xz.sha256'], self.download_directory
```
rjschwei: This breaks for GCE where we create a tar,gz file

mash/utils/web_content.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2017 SUSE Linux GmbH.  All rights reserved.
```
rjschwei: 2018

schaefi:
I'll run some real live tests... stay tuned

schaefi:
I completed my tests, the obs service is functional within the new download concept. The situation that a metadata change shows up while we are downloading is covered only by unit testing. I haven't found an easy way to test this in a more real live scenario.

If there are no objections I'll merge today evening

PR #254, opened 2018-07-18
Added uploader parameters for Azure

### LINKED ISSUE #232: Azure upload performance
Azure upload performance is highly dependent on bandwidth of the originating network and easily falls subject to saturation problems. Therefore we should make the number of connections for image uploading a configuration option such that people on different networks can adjust the value to meet their needs/network conditions.

schaefi:
Azure upload performance is highly dependent on bandwidth of the
originating network. Like in azurectl the performance attributes
with chunk_size and chunk retry should be configurable. This
patch adds the following optional sections to the mash config
file

```yaml
uploader:
  azure:
  # chunk size in bytes, default value taken from azure SDK
  max_chunk_byte_size: 4096

  # max retries on block upload error
  max_chunk_retry_attempts: 5
```

This Fixes #232

### DIFF
--- a/mash/services/uploader/cloud/azure_page_blob.py
+++ b/mash/services/uploader/cloud/azure_page_blob.py
@@ -16,6 +16,7 @@
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
 from collections import OrderedDict
+from mash.services.base_defaults import Defaults
 
 from mash.mash_exceptions import (
     MashAzurePageBlobZeroPageError,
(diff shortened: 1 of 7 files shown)


mash/services/uploader/cloud/azure_page_blob.py
```diff
@@ -53,9 +54,13 @@ def __init__(self, blob_service, blob_name, container, byte_size):
                 '{0}: {1}'.format(type(e).__name__, e)
             )
 
-    def next(self, data_stream, max_chunk_byte_size=None, max_attempts=5):
-        if not max_chunk_byte_size:
-            max_chunk_byte_size = self.blob_service.MAX_CHUNK_GET_SIZE
+    def next(self, data_stream, max_chunk_byte_size=None, max_attempts=None):
```
smarlowucf: Just suggestion but this could be simplified to:

def next(self, data_stream, max_chunk_byte_size=4096, max_attempts=5):

schaefi: I do not understand how it would simplify the method definition ? The goal with the next() method is to allow for calling with:

```python
next(stream)
next(stream, chunk_size)
next(stream, max_attempts=retry_count)
next(stream, chunk_size, retry_count)
```

and in any case a default value should apply it should be taken from the Defaults space
and not be defined somewhere else in the code. I also wanted to avoid to provide the possibility to bypass a method definition default by giving None as the value, e.g

```python
next(stream, max_attempts=None)
```

That's the reason why the parameter setup is done inside of the method. I'm not aware of
preventing the above max_attempts assignment with a method definition default value

Thus I don't see the simplification ?

smarlowucf: Right if you pass None to a kwarg it stomps on the default so I think this is good.

PR #256, opened 2018-07-20
Remove duplicate job doc logging.

smarlowucf:
Job docs are now logged by job creator so only log job id.


smarlowucf:
@rjschwei Good?

PR #257, opened 2018-07-20
Handle special regions for ec2.

smarlowucf:
Accounts in EC2 may have access to special regions. The accounts can now be created with a list of dictionaries to map special region names with helper images.


schaefi:
Thanks then I'm on the right track and all looks good

smarlowucf:
@rjschwei Good?

PR #258, opened 2018-07-20
Remove osc from spec template.

smarlowucf:


smarlowucf:
> The travis error is unrelated to this change. Something in the requires list for the azure sdk on pip seems to have changed

Yeah, and only with older pythons :thinking:

smarlowucf:
Okay problem here is three fold:

- Azure pushed out a ton of updates to packages, some of the subpackages have new versions but the meta packages (azure, azure-mgmt) require old versions.
- In MASH we just require azure which installs everything from top down.
- In IPA I split out only the requirements IPA needs (azure-mgmt-compute, azure-mgmt-network, etc)

IPA installs the latest versions which breaks MASH. But this isn't really an issue with IPA since it should work with latest.

- Should we split out the azure packages we need instead of just requiring all azure?
- Or just add explicit versions to MASH for now:

```
azure-mgmt-compute~=3.0
azure-mgmt-resource~=1.2
```

Just have to make sure these are kept inline with the latest azure meta package requirements.

schaefi:
I think using explicit parts of the sdk in their version as we need it is the best solution. Azure itself has changed the sdk to follow that concept. It's a monster but I think we should adapt. I saw your pull request about it and will comment next. Thanks much

schaefi:
Let's merge this one which fixes the osc requirement to be no longer required

PR #259, opened 2018-07-24
Specific azure requirements versions.

smarlowucf:
Fix build error due to azure version requirements.

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -5,6 +5,10 @@
 # $ pip install -U pip setuptools
 # $ pip install -r .virtualenv.requirements.txt
 
+azure-mgmt-compute
+azure-mgmt-resource
+azure-mgmt-storage
+azure-storage-blob
 setuptools
 idna<2.7
 boto3
@@ -20,5 +24,4 @@ ec2deprecateimg
 ec2publishimg
 ec2uploadimg
 python3-ipa
-azure
 lxml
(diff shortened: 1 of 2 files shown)


.virtualenv.requirements.txt
```diff
@@ -5,6 +5,8 @@
 # $ pip install -U pip setuptools
 # $ pip install -r .virtualenv.requirements.txt
 
+azure-mgmt-compute~=3.0
+azure-mgmt-resource~=1.2
```
schaefi: We also need the storage components

* azure-storage

schaefi:
I also realized the spec file ```package/mash-spec-template``` is missing appropriate requires for the packages. Could you also add:

* python3-azure-sdk

Or do we want to require the component packages:

* python3-azure-mgmt-resource
* python3-azure-mgmt-compute
* python3-azure-storage

Thanks

smarlowucf:
> Or do we want to require the component packages:

I prefer requiring only the components needed.

With azure-storage are we okay with slightly different requirements with spec and pypi or should they match? Azure has storage split up but we don't yet. The component requirement is azure-storage-blob which is in pypi but we only package the meta azure-storage.

schaefi:
>With azure-storage are we okay with slightly different requirements with spec and pypi or should they >match?

yes I think that's ok

smarlowucf:
> yes I think that's ok

Good to merge and fix build?

PR #261, opened 2018-07-27
Add rest api with Flask.

smarlowucf:
Add REST API using Flask:

- Add API and validate endpoints using existing schema.
- Add requirements on Apache and mod_wsgi.
- Add Apache conf file for mash.
- Update spec.
- Integrate API with MASH.
- Move job_id handling to API.
- Remove date-time validation package requirement and use regex instead.

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -9,6 +9,7 @@ azure-mgmt-compute
 azure-mgmt-resource
 azure-mgmt-storage
 azure-storage-blob
+flask
 setuptools
 idna<2.7
 boto3
@@ -18,7 +19,6 @@ PyYAML
 PyJWT
 APScheduler
 python-dateutil>=2.6.0,<2.7.0
-strict-rfc3339
 amqpstorm
 ec2deprecateimg
 ec2publishimg
--- a/config/mash.conf
+++ b/config/mash.conf
@@ -1,0 +1,22 @@
+LoadModule ssl_module modules/mod_ssl.so
+LoadModule wsgi_module modules/mod_wsgi.so
+
+Listen 443
+<VirtualHost *:443>
+    ServerName CHANGE_ME
+
+    WSGIDaemonProcess mash user=mash group=mash threads=5
+    WSGIScriptAlias / /var/lib/mash/wsgi.py
+
+    <Directory /var/lib/mash>
+        WSGIProcessGroup mash
+        WSGIApplicationGroup %{GLOBAL}
+        WSGIScriptReloading On
+        Require ip JUMP_HOST_IP
+    </Directory>
+
+    SSLEngine on
+    SSLCertificateFile /etc/apache2/ssl/server.crt
+    SSLCertificateKeyFile /etc/apache2/ssl/server.key
+    SSLCACertificateFile /etc/apache2/ssl/intermediate.crt
+</VirtualHost>
--- a/package/mash-spec-template
+++ b/package/mash-spec-template
@@ -41,13 +41,15 @@ Requires:       python3-amqpstorm >= 2.4.0
 Requires:       python3-APScheduler >= 3.3.1
 Requires:       python3-python-dateutil >= 2.6.0
 Requires:       python3-python-dateutil < 2.7.0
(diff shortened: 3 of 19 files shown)


schaefi:
puh, that's a big one. Will come back to it when there is more time to review. Is there an easy way to do real live testing ? that might be easier to run through the major code parts. Thanks

smarlowucf:
@schaefi Will work on splitting this up into separate commits. Also, will move this into our service structure. I think it's good as a service but does not need to extend the base_service class.

rjschwei:
Given that we just pushed/ are still pushing images for L1TF, I'd like the focus to be on finishing the rest of the pipeline, i.e. work with James on the new API, which is already happening and then get us to the point where we can put a json file in the "right" place an at least get some of the work done. Endpoints etc. should come later.

smarlowucf:
@schaefi I split this up into three commits. The final commit can be ignored/dropped if we want. It's just a nice way to see the endpoints.

schaefi:
I reviewed https://github.com/SUSE-Enceladus/mash/pull/261/commits/90e3efbc3031d5e750c0df6741e5ab1016e737f8 which looks good to me :+1:

.virtualenv.requirements.txt
```diff
@@ -9,19 +9,20 @@ azure-mgmt-compute
 azure-mgmt-resource
 azure-mgmt-storage
 azure-storage-blob
+flask
+gunicorn
```
schaefi: So we use gunicorn as the http server which is fine by me. Probably I was too busy or missed the communication you had when this decision was made, but I'm fine with it. I think the number of requests and the amount of data this server has to process will not exceed its limits

package/mash-spec-template
```diff
@@ -34,20 +34,21 @@ Requires:       python3-azure-mgmt-storage
...
-Requires:       python3-strict-rfc3339
 Requires:       python3-ec2deprecateimg
 Requires:       python3-ec2publishimg
 Requires:       python3-ec2uploadimg
 Requires:       python3-ipa
 Requires:       python3-ipa-tests
 Requires:       python3-lxml
+Requires:       python3-openapi-core
+Requires:       python3-flask
+Requires:       python3-gunicorn
```
schaefi: We need to make sure sle15 provides them

smarlowucf:
@rjschwei @schaefi This is updated now sticking to existing jsonschema validation for now. It includes the requirements for apache + mod_wsgi. Also, a template mash conf for Apache.

Thoughts on auth for now? I think a good option is basic auth in Apache using .htaccess file. Or if Apache can be configured to use PAM and the existing SSH keys?

config/mash.conf
```diff
@@ -0,0 +1,14 @@
...
+
+    WSGIDaemonProcess mash user=mash group=mash threads=5
+    WSGIScriptAlias / /var/lib/mash/wsgi.py
+
+    <Directory /var/lib/mash>
+        WSGIProcessGroup mash
+        WSGIApplicationGroup %{GLOBAL}
+        WSGIScriptReloading On
+        Order deny,allow
+        Allow from all
```
rjschwei: The deny,allow syntax is Apache 2.2.hold over syntax and is deprecated, we should use the new directives

https://httpd.apache.org/docs/2.4/howto/access.html

rjschwei:
For auth for the initial setup lets allocate an IP in AWS in the prod account and create a jump host. The host can have our ssh keys, plus James and the apache config contains

<Require all>
   ip THE_ALLOCATED_IP
</Require all>

That way we don't have to fiddle with passwords, auth setup etc. Jobs can only be submitted from the jump host and only we can login.

config/mash.conf
smarlowucf: Template updated to require ip of jump host.

smarlowucf:
> For auth for the initial setup lets allocate an IP in AWS in the prod account and create a jump host.

Sounds good.

smarlowucf:
@schaefi The first commit https://github.com/SUSE-Enceladus/mash/pull/261/commits/8b9737192a3863b5e27f59be064cfa2b8f674a85 has been updated to use the existing validation schema and thus much smaller changeset.

PR #263, opened 2018-08-16
Use the JsonFormat class for json dumps.

smarlowucf:
- Ensure all json message strings have the same format.
- Make JsonFormat methods static as they make no use of the uninstantiated class.

### DIFF
--- a/mash/services/base_service.py
+++ b/mash/services/base_service.py
@@ -36,6 +36,7 @@ from mash.mash_exceptions import (
     MashLogSetupException,
     MashValidationException
 )
+from mash.utils.json_format import JsonFormat
 
 
 class BaseService(object):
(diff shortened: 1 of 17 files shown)


schaefi:
Thanks, this is a good change. I was already wondering why nobody other than me uses my little JsonFormat class :)

I'm good with the change to @staticmethod, actually that really makes sense as the method does not share any data in the scope of that namespace. So we don't need an instance of JsonFormat which we implicitly still have when implementing it as @classmethod

So all good thanks :+1:

smarlowucf:
> I was already wondering why nobody other than me uses my little JsonFormat class :)

I hadn't used it initially since I was just using json.dumps() with no args so it didn't make sense to import a class to use a one line method that replaces the same line.

But given we need sort_keys and use indent it makes sense to use have a shared method to encapsulate this.

PR #274, opened 2018-08-31
Send status message on obs job failure

schaefi:
For not nonstop jobs and in case of an error a status message
must be send to allow cleanup of the jobs in the service
queue referencing this job

### DIFF
--- a/mash/services/obs/build_result.py
+++ b/mash/services/obs/build_result.py
@@ -348,6 +348,8 @@ class OBSImageBuildResult(object):
                 self._log_callback('Job done')
                 self._result_callback()
         except MashException as issue:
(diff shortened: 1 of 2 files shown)


schaefi:
on a short break from vacation I noticed the message from Sean with the report on the obs service.
Thanks Sean, this one should fix the first issue you reported.

schaefi:
Seam, I also looked at the uploader code which does not have the issue fixed here for obs. So  uploader sends status messages on error

schaefi:
I'll be leaving on the second part of the hiking tour today, thus will be out with no network next week again :)

rjschwei:
LGTM

PR #286, opened 2018-09-12
Match the image name for downloading on minimum match criteria

rjschwei:
+ At present only x86_64 is supported for cloud images, thus the
    architecture in a name is not a distinguishing feature between
    images such as xyz.x86_64 and xyz-abc.x86_64 The distinguishing feature
    in a glob is the '.' after the image name, not the presence of the
    architecture designation.

### DIFF
--- a/mash/services/obs/build_result.py
+++ b/mash/services/obs/build_result.py
@@ -73,9 +73,6 @@ class OBSImageBuildResult(object):
           {'image': '1.42.1'}
       ]
 
-    * :attr:`arch`
-      Buildservice package architecture, defaults to: x86_64
-
     * :attr:`download_root`
       Buildservice URL, defaults to: https://api.opensuse.org
 
@@ -86,8 +83,7 @@ class OBSImageBuildResult(object):
       Download directory name, defaults to: /tmp
     """
     def __init__(
-        self, job_id, job_file, project, image_name,
-        conditions=None, arch='x86_64',
+        self, job_id, job_file, project, image_name, conditions=None,
         download_root='https://download.opensuse.org/repositories',
         repository='images', download_directory=Defaults.get_download_dir()
     ):
@@ -96,7 +92,6 @@ class OBSImageBuildResult(object):
         self.download_directory = download_directory
         self.download_root = download_root
         self.repository = repository
-        self.arch = arch
         self.project = project
         self.image_name = image_name
         self.image_metadata_name = None
@@ -208,7 +203,7 @@ class OBSImageBuildResult(object):
 
     def _get_build_number(self, name):
         build = re.search(
-            '{0}-(.*)-Build([0-9]+\.[0-9]+)'.format(self.arch), name
+            '-(\d+\.\d+\.\d+)-Build([0-9]+\.[0-9]+)', name
         )
         if build:
             return [build.group(1), build.group(2)]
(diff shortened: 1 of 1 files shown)


schaefi:
Hmm, the arch flag is set from an initialize option on constructing a OBSImageBuildResult instance. If not set it defaults to x86_64 which is how we construct it at the moment. I think if the instance is created for an architecture the match criteria should align to it.

I have concerns to weaken this with the suggested change and would agree to it only if you are running into matching problems which I don't see

schaefi:
and to mention any kiwi created image is for a specific architecture and you can rely on the name to contain an arch information. For me this is most important if we ever want to support other than the x86_64 architecture

schaefi:
> When/if cloud providers start supporting other architectures then yes the architecture becomes a
> distinguishing part of the name. Until then it is not and thus falls under the category YAGNI

good, I can live with that, but then the change is incomplete. You don't want any distinguish by arch then it should be deleted in a clean way for this class. I use the arch selector in the following places:

* watching for the results in _watch_obs_result
* when fetching the image in get_image
* when extracting the build number in _get_build_number
* and of course in the constructor as __init__: arch='x86_64'

I have tested all that code with an arch selector because I know this will be present 100% in any of the images kiwi builds. As I own kiwi I can guarantee it. There is also Sean's last commit #30cb4ecab6b98c79 which fixes the fetcher by an arch selector

Last but not least you still did not tell me what issue in fetching the images you had with the arch selector in place. To me this feels like a unnecessary change in the middle of a road to success introducing potential  errors because now I don't know for sure if the match still will work

So why ?

schaefi:
I said I can live with the change if we complete it. I will do the necessary bits and commit shortly

>Please write a regular expression that differentiates 'xyz.x86_64' and 'xyz-abc.x86_64', let's see if you >need the architecture, 

The match from the index list is based on base_name which needs to match ```startswith``` and a suffix_list which needs to match ```endswith```. For me its obvious that an image has an architecture and that should be part of the base_name.

We can easily go without that automatic base_name extension and handle it in the job document where
we say

```
{
    "obs_job": {
        "id": "123",
        "image": "OEM-openSUSE-Tumbleweed",
```

we could also say

```
"image": "OEM-openSUSE-Tumbleweed.x86_64",
```

or even more fine grained.

schaefi:
The last commit completes the removal of arch handling.

Please note as soon as a project url is used which for whatever reason has image builds for multiple archs published our obs service will match all of them and could potentially download an image for the wrong architecture and send that along the service queue. As of today this will not happen but we will get notified only by accident because without the arch information there is no "error" source available anymore

Open for review again

Thanks

rjschwei:
Thanks, LGTM

Your concerns on the architecture are valid and noted. If we ever get to the point where something else than x86_64 is supported/needed in the Public Cloud we have to have a discussion about that. We will have to figure out if it would make sense to treat any given architecture in a special way.

For example if we need to release image X on architecture A for a security issue it is very likely that the same image would also need to be released on architecture B. This would imply that we might not want to call out the architecture specifically and just let mash handle all the architectures automatically. On the other hand if we do decide that architecture specific image releases rather than an "all architecture" approach makes sense then the architecture would have to be exposed in the job document and that requires more changes then just these. In that case the change set should go in as one consistent commit, IMHO.

Thanks again for picking up after me.

smarlowucf:
@schaefi Looks good. I have another change needed to handle beta build numbers. https://github.com/SUSE-Enceladus/mash/compare/obs-build-number?expand=1

Can do a separate PR and merge and cherry-pick commit here?

schaefi:
@smarlowucf yes looks good to me, just commit the change of the beta build numbers here and I'll wait for you before I merge. Thanks

smarlowucf:
Going ahead with merge. Beta number changes included.

PR #288, opened 2018-09-13
Combine OBS project and download_root.

smarlowucf:
Into a single required argument download_url. And remove old references to api_url.


schaefi:
This probably creates a merge conflict with #286, I suggest to merge in #286 first after you have added the beta build number commit and then resolve this one if there is a conflict. Thanks

PR #289, opened 2018-09-18
If no instance type provided select random.

smarlowucf:
Test against a random instance type if type is None.

@rjschwei  I think we should have a small subset of instance types per cloud. Are these lists okay? Any additions or deletions needed?

Also if we are testing against random regions then we need to sure that the instance types in list exist for all regions.

Also, when testing against very large instances in certain regions the provision and reboot time may be very slow. So we may need to override ipa timeout option with a very high value (default is 10 minutes per state change).

### DIFF
--- a/mash/services/testing/azure_job.py
+++ b/mash/services/testing/azure_job.py
@@ -26,8 +26,11 @@ from mash.services.testing.ipa_helper import ipa_test
 from mash.services.testing.job import TestingJob
 
 instance_types = [
-    'Standard_DS1_v2', 'Standard_B1ms', 'Standard_F1s', 'Standard_H8m',
-    'Standard_A0', 'Standard_D1_v2', 'Standard_A2m_v2'
+    'Basic_A2',
+    'Standard_B1s',
+    'Standard_D2_v3',
+    'Standard_E2_v3',
+    'Standard_F2s_v2'
 ]
 
 
--- a/mash/services/testing/azure_job.py
+++ b/mash/services/testing/azure_job.py
@@ -17,12 +17,18 @@
 #
 
 import json
+import random
 
 from threading import Thread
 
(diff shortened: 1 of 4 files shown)


mash/services/testing/ec2_job.py
```diff
@@ -16,12 +16,19 @@
...
 
 from mash.services.status_levels import FAILED, SUCCESS
 from mash.services.testing.ipa_helper import ipa_test
 from mash.services.testing.job import TestingJob
 
+instance_types = [
+    't2.micro', 'm5.large', 'c5d.large', 'x1e.xlarge',
+    'p3.2xlarge', 'h1.2xlarge', 'd2.xlarge'
+]
+
```
rjschwei: Pleas add:

r5.24xlarge, i3.metal, i3.8xlarge, d2.xlarge, t3.small, m5d.large

this should give us an instance type for every network config supported.

mash/services/testing/azure_job.py
```diff
@@ -17,13 +17,19 @@
...
+import random
 
 from threading import Thread
 
 from mash.services.status_levels import FAILED, SUCCESS
 from mash.services.testing.ipa_helper import ipa_test
 from mash.services.testing.job import TestingJob
 
+instance_types = [
+    'Standard_DS1_v2', 'Standard_B1ms', 'Standard_F1s', 'Standard_H8m',
```
bear454: I'd recommend the following types for general testing:

* Standard_B1s
* Standard_D2_v3
* Basic_A2 (Basic disk)
* Standard_F2s_v2 (Compute optimized)
* Standard_E2_v3 (Memory optimized)

And for HPC:
* Standard_A8
* Standard_H16r

smarlowucf:
Thanks. Lists updated. Azure HPC instance types are handled in job docs.

PR #302, opened 2018-09-21
Remove utctime handling from Uploader service.

### LINKED ISSUE #297: Uploader service fails handling utctime based jobs
Uploader service attempts to handle utctimes and schedules jobs based on these times. This causes the job to fail. No matter if OBS send status message for the job it never starts.

Likely cause is the job gets scheduled based on provided utctime. However, that time passes before the uploader service can move forward. Thus the job in background scheduler is stuck.

Instead the uploader service should treat utctime base jobs just like now jobs. OBS service handles the scheduling of utctime jobs. Uploader should just wait to receive a message from OBS service.

smarlowucf:
Only OBS service should handle specific utctimes. All other services only distinguish between always and now/utctime jobs.

Fixes #297

### DIFF
--- a/mash/services/uploader/service.py
+++ b/mash/services/uploader/service.py
@@ -18,7 +18,6 @@
 import atexit
 import os
 import time
-import dateutil.parser
 from pytz import utc
 from apscheduler.schedulers.background import BackgroundScheduler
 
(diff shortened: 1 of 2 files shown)


rjschwei:
The obs service handles all of this. When the obs service decides that an image is downloaded and meets all the conditions then the rest of the pipeline should just proceed. I don't recall a discussion where we'd need the uploader to be specific about the time. Also, the uploader job-doc would have the same utctime as the obs job-doc, thus by the time the uploader get the "proceed" message from the obs service that time by definition would have already passed.

smarlowucf:
> Also, the uploader job-doc would have the same utctime as the obs job-doc, thus by the time the uploader get the "proceed" message from the obs service that time by definition would have already passed.

Right, and the test I ran yesterday proved this to be a problem anyway. Because the job was scheduled based on same time as OBS. Thus when uploader got the image and could move forward the background scheduler was stuck because the job was marked with a time in the past.

PR #305, opened 2018-09-21
Add schema for GCE job and account messages.

smarlowucf:


smarlowucf:
Rebase accounting for bucket in schema.

PR #306, opened 2018-09-21
Add GCE handling to job creator.

smarlowucf:


smarlowucf:
Rebase accounting for bucket in schema.

mash/services/jobcreator/gce_job.py
```diff
@@ -0,0 +1,132 @@
...
+    def get_testing_regions(self):
+        """
+        Return a dictionary of target test regions.
+        """
+        test_regions = {}
+
+        for source_region, value in self.target_account_info.items():
+            test_regions[source_region] = value['account']
+
+        return test_regions
```
rjschwei: For testing, i.e. runing an instance we need zone information. Somewhere we will need to make the jump between a region us-east1 and a zone us-east1-b

mash/services/jobcreator/gce_job.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _get_account_info(self):
+        """
+        Returns a dictionary of regions to accounts.
+
+        Example: {
+            'us-west1': {
+                'account': 'acnt1',
+                'bucket': 'images'
+            }
```
rjschwei: There are no disconnected regions in GCE at this time, thus the same account applies to all regions. But we do want to be prepared in case disconnected regions do show up in GCE.

mash/services/jobcreator/gce_job.py
smarlowucf: I think we'd just use the existing region functionality for the zone. Region can be extracted from zone if necessary. I.e. in testing service for GCE both zone and region are both used https://github.com/SUSE-Enceladus/ipa/blob/master/ipa/ipa_gce.py#L193.

mash/services/jobcreator/gce_job.py
smarlowucf: Right only needed in testing service. But I think it's nice to maintain a similar flow as the other providers.

PR #308, opened 2018-09-24
Share with and allow copy are optional.

smarlowucf:
If not provided share with default is `all` and allow copy is `False`.

### DIFF
--- a/mash/services/jobcreator/ec2_job.py
+++ b/mash/services/jobcreator/ec2_job.py
@@ -32,8 +32,9 @@ class EC2Job(BaseJob):
         self, accounts_info, provider_data, job_id, provider,
         provider_accounts, provider_groups, requesting_user, last_service,
         utctime, image, cloud_image_name, old_cloud_image_name,
-        share_with, allow_copy, image_description, distro, download_url,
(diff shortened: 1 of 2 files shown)


mash/services/jobcreator/ec2_job.py
```diff
@@ -32,8 +32,9 @@ def __init__(
         self, accounts_info, provider_data, job_id, provider,
         provider_accounts, provider_groups, requesting_user, last_service,
         utctime, image, cloud_image_name, old_cloud_image_name,
-        share_with, allow_copy, image_description, distro, download_url,
-        tests, conditions=None, instance_type=None
+        image_description, distro, download_url, tests,
+        allow_copy=False, conditions=None, instance_type=None,
+        share_with='all'
```
rjschwei: This is too risky, the default for sharing should be 'None' unless this is required in the job doc. Having the default 'all' makes it too easy for mistakes to happen. I rather run a manual publishing then the revers.

smarlowucf: Okay, just following the defaults in the ec2publishimg utility.

smarlowucf:
Amended commit with default sharing == 'none'

PR #311, opened 2018-09-26
Add family option to GCE jobs.

smarlowucf:
Family is required by uploader service when creating image.

### DIFF
--- a/examples/messages/gce_job.json
+++ b/examples/messages/gce_job.json
@@ -19,5 +19,6 @@
   "image_description": "New Image #123",
   "distro": "sles",
   "instance_type": "t2.micro",
+  "family": "sles-15",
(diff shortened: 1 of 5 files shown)


examples/messages/gce_job.json
```diff
@@ -19,5 +19,6 @@
   "image_description": "New Image #123",
   "distro": "sles",
   "instance_type": "t2.micro",
+  "family": "sles-11",
```
rjschwei: Since we do not support SLES 11 in mash for GCE using this as a family in the "example" is probably not the best choice.

smarlowucf: :smile:

smarlowucf: Rebased with sles-15

PR #312, opened 2018-09-26
Add GCE job handling in testing service.

smarlowucf:


mash/services/testing/gce_job.py
```diff
@@ -0,0 +1,75 @@
...
+    'f1-micro',
+    'n1-ultramem-40'
+]
+
+
+class GCETestingJob(TestingJob):
+    """
+    Class for an GCE testing job.
+    """
+    __test__ = False
```
schaefi: this global class variable appears once, what is the purpose ?

rjschwei: Also exists in all the other implementations but does not appear to be used. Also no reference found in the ipa code

smarlowucf: Cannot find this in the docs but here is the pytest code which checks for the `__test__` flag https://github.com/pytest-dev/pytest/blob/49800ea13434ee08e2d8e5b1d1b7a3fb4061b4af/src/_pytest/unittest.py#L47.

smarlowucf: Also, this comment references the fact that at some point subclasses also needed the flag to prevent discovery. https://github.com/pytest-dev/pytest/issues/1879#issuecomment-243241623.

rjschwei:
Ah yes, pytest looks for "test" in a name and when you set pytest loose on these files it will probably get confused and think the implementation is a test. The __test__ = False prevents such mistakes.

Lets see if we can get away without it and if not we'll re-add it, simple enough, with a comment to remind us later.

smarlowucf:
Yes, I've removed it from the subclasses and it doesn't cause problems. It remains in the base job.py module. Will submit a separate PR to add a comment which will make the usage clear.

PR #314, opened 2018-10-01
Delete job doc logging from obs and uploader

### LINKED ISSUE #276: Services echo job doc info
The OBS, Uploader and Credentials service echo the job doc in log. These should be removed as the job creator service already prints the entire job doc.

- [x] Remove job doc print in credentials log
- [x] Remove job doc print in obs log
- [x] Remove job doc print in uploader log

schaefi:
The complete job documents are logged by the job creator
service. Thus there is no need for other services to log
their individual parts of it. This Fixes #276


rjschwei:
LGTM

PR #315, opened 2018-10-01
Temporary type changes in EC2 job.

smarlowucf:
Libcloud is far behind on the EC2 types available. Thus the list must match what Libcloud has until a patch can be pushed through.


smarlowucf:
Closing this for now. Patched libcloud will be used with the existing list.

PR #323, opened 2018-10-05
Remove service based configs from testing data.

### LINKED ISSUE #321: Cleanup leftover service based config examples
MASH config is a single YAML with sections for services. There are still service based yaml examples in code base. Cleanup these old examples to reduce confusion.

smarlowucf:
- Use a single config in testing to reduce confusion.
- Update config class docstrings to point to a single mash config.

Fixes #321


mash/services/logger/config.py
```diff
@@ -23,13 +23,12 @@
 
 class LoggerConfig(BaseConfig):
     """
-    Implements reading of logger service configuration file:
+    Implements reading of mash configuration file:
```
rjschwei: I think we want a more explicit comment. IMHO it's confusing to have "class LoggerConfig" followed by "...of mash configuration file". I think a comment like:

"Implements reading of logger configuration from mash configuration file:"

is more explanatory.

smarlowucf:
Updated docstrings for all service config classes.

PR #324, opened 2018-10-05
Update accounts file location in config.

### LINKED ISSUE #320: Credentials accounts_file in wrong section
With the switch from account handling to credentials service the accounts_file in config was not moved to credentials section.

https://github.com/SUSE-Enceladus/mash/blob/master/mash/services/credentials/config.py#L44

Move this value to credentials section and cleanup example config files.

smarlowucf:
- Move to credentials section.
- Update example mash_config.yaml with new location.

Resolves #320

### DIFF
--- a/config/mash_config.yaml
+++ b/config/mash_config.yaml
@@ -18,7 +18,6 @@ non_cred_services:
   - obs
 credentials:
   credentials_directory: /var/lib/mash/credentials/
-jobcreator:
   accounts_file: /etc/mash/accounts.json
 obs:
   download_directory: /var/lib/mash/images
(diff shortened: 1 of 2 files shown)


rjschwei:
I also came to the conclusion that there were some unrelated changes, but didn't think they were significant enough to make this PR be separated.

The change to SUSE LLC is based on the information from [1]

"....Use “SUSE LLC” when attributing the copyright. Use your “@suse.com” email address for contributions."

We should slowly migrate to this when we touch files anyway.

[1] https://opensource.suse.com/suse-open-source-policy

smarlowucf:
As @rjschwei mentioned this was just a change to follow the open source policy. Given I was touching the modules I made the change where necessary.

Since it's a none code change I pushed in same commit. But understand it'd be less confusing in separate commit.

PR #328, opened 2018-10-23
Integrate Azure replication

smarlowucf:
This is a bit to review but it's broken down into multiple commits.

- Add azure utils for handling the copy of arm blob to legacy storage.
- Update azure schema with source (arm) and destination (legacy) storage info.
- Integrate replication for azure.
- Add credentials endpoint urls to azure schema validation.
- Integrate page blob deletion.
- Integrate arm image deletion.

The last two would end up as part of the image cleanup service when it's integrated. That way they could be called if testing only. But I think it's good to have them here for now.

@bear454 If you want you can take a look at this as well. Is there anything missing?

### DIFF
--- a/mash/services/uploader/cloud/azure.py
+++ b/mash/services/uploader/cloud/azure.py
@@ -42,7 +42,7 @@ class UploadAzure(UploadBase):
 
         custom_args={
             'region': 'region_name',
-            'container_name': 'storage_container_name',
+            'container': 'storage_container_name',
             'storage_account': 'storage_account_name',
             'resource_group': 'optional_resource_group_name'
         }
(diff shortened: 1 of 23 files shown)


mash/services/uploader/cloud/azure.py
```diff
@@ -57,8 +57,8 @@ def post_init(self):
                 'required Azure region name for upload not specified'
             )
 
-        self.container_name = self.custom_args.get('container_name')
-        if not self.container_name:
+        self.container = self.custom_args.get('container')
+        if not self.container:
```
rjschwei: This would also be inconsistent. The message says that the "container name" was not found, but the test implies that we are handling a container object.

smarlowucf: The usage for all args matches this. I.e. resource_group and storage_account. So this make things consistent. All of them are name strings not objects. So if we want to change this back to name I'd prefer in a separate PR where everything changes:

container -> container_name
resource_group -> resource_group_name
storage_account -> storage_account_name

rjschwei: Fair

PR #329, opened 2018-10-24
Re-raise exceptions in event loop.

### LINKED ISSUE #325: Uploader service fails silently
If there is an error that occurs in the event loop while consuming messages the uploader service catches the exception and stops consuming. However, the exception is never re-raised so the service fails silently with no info on what caused the exception.

Update service to re-raise exceptions after the rabbitmq connections are closed.

smarlowucf:
Always close connections when event loop exits.

Fixes #325

### DIFF
--- a/mash/services/uploader/service.py
+++ b/mash/services/uploader/service.py
@@ -63,9 +63,9 @@ class UploadImageService(BaseService):
         try:
             self.channel.start_consuming()
         except Exception:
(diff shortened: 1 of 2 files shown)


rjschwei:
Tests failing

smarlowucf:
> Tests failing

See #330. Fixes new warnings from updated flake8 package.

PR #333, opened 2018-10-25
Use ARM endpoint for Azure upload 

### LINKED ISSUE #327: Uploader very slow for azure
The Azure upload is extremely without the use of multiple workers. In testing an image can take over 10 hours to upload.

To speed up the upload process either directly use the create_blob_from_stream method from azure sdk or use concurrent futures and threading to copy a number of pages (max_workers) at a time.

smarlowucf:
Instead of using page_upload module use ARM create_blob_from_stream method. This allows upload across multiple workers for better performance.

Fixes #327


mash/services/uploader/cloud/azure.py
```diff
@@ -90,24 +89,33 @@ def upload(self):
...
+        while True:
+            with open_image(self.system_image_file, 'rb') as image_stream:
+                try:
+                    page_blob_service.create_blob_from_stream(
+                        self.container_name, blob_name, image_stream,
+                        system_image_file_type.get_size(),
+                        max_connections=self.config.get_azure_max_workers()
+                    )
+                    break
+                except Exception as e:
```
schaefi: Can we give exception variables a more meaningful name, e.g issue or error not just e ?

smarlowucf: Yes, this is a tricky one since we want the error message out of scope if it's raised below. However, exception and stacktrace are cleared out when you exit the exception block.

smarlowucf: Re-based using `error` and `msg` to capture the value for outer scope.

Issue #298, opened 2018-09-20
OBS service not sending delete job message into pipeline


smarlowucf:
If obs receives a delete job message it deletes the job and logs such. However, it never send this along to uploader service to clean out the pipeline.

OBS should send a failure status message with status == 'delete' to uploader service.

PR #337, opened 2018-11-02
Properly handle job if failed upstream.

smarlowucf:
Uploader should set job to ready so the scheduled job can finish. The start job method will delete job and send message downstream.

@schaefi This is more fixes related to https://github.com/SUSE-Enceladus/mash/issues/298.

I didn't fully get this patched correctly. In that the message was sent and uploader would attempt to delete job but failed because the scheduled job was stuck waiting on ready status.

So I modified the handling to set ready and a failed flag. This is caught in the start job method where the job is deleted and message sent to testing service.

Did I catch everything this time?

### DIFF
--- a/mash/services/uploader/service.py
+++ b/mash/services/uploader/service.py
@@ -151,18 +151,8 @@ class UploadImageService(BaseService):
                 )
                 self._check_ready(job_id)
             else:
-                result = self._delete_job(job_id)
-                self._send_control_response(result, job_id)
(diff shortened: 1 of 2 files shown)


schaefi:
Per conversation on e-mail I think this one is replaced by #345 

I set the block label such that it does not get merged accidentally

smarlowucf:
Yeah this is covered by #345. Closing.

PR #348, opened 2018-11-13
Pass job last_service to all services

### LINKED ISSUE #341: Credential service cleanup
If a job fails before or during the final service the credentials service doesn't know to cleanup the job. Thus old failed jobs start to pile up in the credentials service.

There needs to be some way to ping credentials service to cleanup these jobs.

### LINKED ISSUE #346: No result message on last service
MASH services currently will always send the job results to the service in the pipeline. If the job doc for  the last service contained a last_service flag then MASH would know to not publish a message.

This will cleanup some of the noise in the message queue and for downstream services which end up processing these messages only to toss them out.

This is also needed for #341 to allow the last service to notify credentials of a failed job to cleanup.

smarlowucf:
This is required to prevent extra noise when a job has finished. If the job's last service is the current service don't publish a message downstream.

last_service is also needed to fix 341 .

Resolves #346

### DIFF
--- a/examples/messages/obs_always_job.json
+++ b/examples/messages/obs_always_job.json
@@ -3,6 +3,7 @@
         "id": "5678",
         "download_url": "http://download.suse.de/ibs/Devel:/PubCloud:/Stable:/Images12/images",
         "image": "SLES12-Azure-BYOS",
+        "last_service": "obs",
         "utctime": "always",
         "conditions": [
             {"package": ["kernel-default", ">=3.12.61", ">=52.136"]},
--- a/examples/messages/obs_always_job.json
+++ b/examples/messages/obs_always_job.json
@@ -3,7 +3,7 @@
         "id": "5678",
         "download_url": "http://download.suse.de/ibs/Devel:/PubCloud:/Stable:/Images12/images",
         "image": "SLES12-Azure-BYOS",
-        "last_service": "obs",
+        "last_service": "testing",
         "utctime": "always",
         "conditions": [
             {"package": ["kernel-default", ">=3.12.61", ">=52.136"]},
--- a/examples/messages/obs_job.json
+++ b/examples/messages/obs_job.json
@@ -3,7 +3,7 @@
         "id": "123",
         "image": "openSUSE-Tumbleweed-JeOS",
         "download_url": "http://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/openSUSE-Tumbleweed/openSUSE_Tumbleweed",
-        "last_service": "obs",
+        "last_service": "testing",
         "utctime": "now",
         "conditions": [
             {"package": ["kernel-default", ">=4.4.1", ">=1.1"]},
--- a/examples/messages/obs_job.json
+++ b/examples/messages/obs_job.json
@@ -3,6 +3,7 @@
         "id": "123",
         "image": "openSUSE-Tumbleweed-JeOS",
         "download_url": "http://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/openSUSE-Tumbleweed/openSUSE_Tumbleweed",
+        "last_service": "obs",
         "utctime": "now",
         "conditions": [
             {"package": ["kernel-default", ">=4.4.1", ">=1.1"]},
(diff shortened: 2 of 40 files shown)


rjschwei:
I disagree that this is needed to address #341, there should be a separate message to address #341

rjschwei:
I don't think any of the services need to know about "last_service". The job doc contains the information and thus the job creator, having knowledge of the order of the pipeline should push the job to only those services that are expected to run.

In that case the last_services that finishes up continues to pass along the job as usual and the next service simply says "I have no job for this ID" and so on. With this approach non of the services have to know if they are "special" and are the last service for a given job. The only service that knows about things being special is the jobcreator.

I don't think we want to take the direction implemented here.

smarlowucf:
I guess I'm less inclined to prefer message noise. Also the current approach given the messaging flow has essentially a memory leak that requires manual intervention.

I.e. if we have a testing job it will force push a message (which includes creating a binding) to the replication service. The replication service will get the message, find no matching job, and drop the message. However, it won't cleanup the binding. So these bindings will build up.

schaefi:
I think the other design refactoring topics:

* handshake between services and credentials service
* use single routing key instead of job_id as key which requires bind/unbind per job

Will be done in separate pull requests ?

So I reviewed this one only for the scope of adding the last_service information to jobs

smarlowucf:
> Will be done in separate pull requests ?

Yes, they would be separate.

examples/messages/obs_always_job.json
```diff
@@ -3,6 +3,7 @@
         "id": "5678",
         "download_url": "http://download.suse.de/ibs/Devel:/PubCloud:/Stable:/Images12/images",
         "image": "SLES12-Azure-BYOS",
+        "last_service": "obs",
```
rjschwei: We agreed that obs cannot be the last service. Therefore our example cannot show this.

examples/messages/obs_job.json
```diff
@@ -3,6 +3,7 @@
         "id": "123",
         "image": "openSUSE-Tumbleweed-JeOS",
         "download_url": "http://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/openSUSE-Tumbleweed/openSUSE_Tumbleweed",
+        "last_service": "obs",
```
rjschwei: obs cannot be the last service as agreed via e-mail.

examples/messages/obs_now_job.json
```diff
@@ -3,6 +3,7 @@
         "id": "0815",
         "download_url": "http://download.suse.de/ibs/Devel:/PubCloud:/Stable:/Images12/images",
         "image": "SLES12-Azure-BYOS",
+        "last_service": "obs",
```
rjschwei: dito

mash/services/jobcreator/schema.py
```diff
@@ -206,7 +206,7 @@
         'requesting_user': {'$ref': '#/definitions/non_empty_string'},
         'last_service': {
             'enum': [
-                'obs', 'uploader', 'testing', 'replication',
+                'uploader', 'testing', 'replication',
```
rjschwei: And here obs is removed as option for last_service, as expected.

mash/services/deprecation/service.py
```diff
@@ -295,8 +297,10 @@ def _process_deprecation_result(self, event):
                 extra=metata
             )
 
-        # Don't send failure messages for always jobs.
-        if job.utctime != 'always' or job.status == SUCCESS:
+        # Don't send failure messages for always jobs and
+        # don't send message if last service.
+        if (job.utctime != 'always' or job.status == SUCCESS) \
+                and job.last_service != self.service_exchange:
```
rjschwei: All services should behave the same, this should be in the base class.

mash/services/deprecation/service.py
```diff
@@ -98,7 +98,9 @@ def _cleanup_job(self, job, status):
             pass
 
         self._delete_job(job.id)
-        self._publish_message(job)
+
+        if job.last_service != self.service_exchange:
+            self._publish_message(job)
```
rjschwei: All services should behave the same, this should be in the base class.

mash/services/deprecation/service.py
smarlowucf: There's a bit more messaging functionality in each service that is not yet shared in base class. I think we should split out the aggregation to a separate PR.

smarlowucf:
Rebased against master to pull in updated ec2imgutils requirements.

smarlowucf:
Ref #351

PR #354, opened 2018-12-07
Use temporary subnet and security group.

smarlowucf:
This guarantees the test instance is accessible on ssh port.


mash/services/testing/ipa_helper.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 SUSE Linux GmbH.  All rights reserved.
+# Copyright (c) 2018 SUSE LLC.  All rights reserved.
```
rjschwei: Good catch, thank you

smarlowucf:
Updated IPA version requirement.

PR #362, opened 2018-12-13
Add next service attribute to services.

smarlowucf:
This removes hard coded service names from classes and is a step to share messaging code.

### DIFF
--- a/mash/services/base_service.py
+++ b/mash/services/base_service.py
@@ -51,7 +51,7 @@ class BaseService(object):
     * :attr:`service_exchange`
       Name of service exchange
     """
-    def __init__(self, service_exchange, next_service=None):
+    def __init__(self, service_exchange):
         self.channel = None
         self.connection = None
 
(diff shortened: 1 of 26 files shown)


rjschwei:
Since you are going through the trouble to refactor this why not get to the end result in one step as it requires probably an equal amount of work.

We have the service order in config/mash_config.yaml so instead of hardcoding into "uploader" that the next service is "testing" each service should call _set_next_service in the c-tor and _set_next_service is implemented in the base class and reads the config and based on self.service_name finds the next service from the config and sets self.next_service

Sorry I a missing where the hardcoding approach in this patch set makes the changeset smaller compared to the more general approach.

smarlowucf:
End goal to change:

`publish_stuff(stuff, 'deprecation')`

to

`publish_stuff(stuff, self.last_service)`

because the latter can move to base service.

smarlowucf:
Updated to use service list.

smarlowucf:
Rebased to fix merge conflicts.

PR #367, opened 2018-12-18
Add prev service to base service class.

smarlowucf:
Used in sharing messaging methods.


rjschwei:
The code is fine, but I am not certain of the use case. We do not really want any backwards communication.

PR #370, opened 2018-12-19
 Rename base_service to mash_service.

smarlowucf:
And BaseService to MashService.

Update copyright notices.

### DIFF
--- a/mash/services/logger/service.py
+++ b/mash/services/logger/service.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017 SUSE Linux GmbH.  All rights reserved.
+# Copyright (c) 2018 SUSE LLC.  All rights reserved.
 #
 # This file is part of mash.
 #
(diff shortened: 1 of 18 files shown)


mash/services/logger/service.py
```diff
@@ -19,10 +19,10 @@
 import json
 
 from mash.mash_exceptions import MashLoggerException
-from mash.services.base_service import BaseService
+from mash.services.mash_service import MashService
 
 
-class LoggerService(BaseService):
+class LoggerService(MashService):
```
rjschwei: Shouldn't this be a PipelineService now? similar for others?

smarlowucf: No LoggerService isn't in the pipeline. It's more of an auxiliary service like Credentials and JobCreator. So it doesn't share all of the listener message methods that are in PipelineService.

mash/services/uploader/service.py
```diff
@@ -32,7 +32,7 @@
 )
 
 
-class UploadImageService(BaseService):
+class UploadImageService(MashService):
```
rjschwei: PipelineService?

smarlowucf:
Rebased with master to fix merge conflicts.

PR #377, opened 2019-01-24
Add architecture option to support aarch64.

smarlowucf:


smarlowucf:
> For AWS you have to do a translation
> 
> if self.arch == 'aarch64':
> aws_arch = 'arm64'
> 
> I didn't see this, sorry if I missed it during review.

Do we instead just have a third option: `['x86_64', 'aarch64', 'arm64']`?

rjschwei:
> > For AWS you have to do a translation
> > if self.arch == 'aarch64':
> > aws_arch = 'arm64'
> > I didn't see this, sorry if I missed it during review.
> 
> Do we instead just have a third option: `['x86_64', 'aarch64', 'arm64']`?
> 
> Or
> 
> This is initially just for the new EC2 arch so swap arm for aarch in schema.

We need "aarch64" for looking up the image name, but for image creation in EC2 it needs to be "arm64". I don't see how we get around the mapping. If we allow "arm64" as a value for arch then that will not work with finding the image from the build service.

smarlowucf:
> We need "aarch64" for looking up the image name, but for image creation in EC2 it needs to be "arm64". I don't see how we get around the mapping. If we allow "arm64" as a value for arch then that will not work with finding the image from the build service.

I figured they could be swapped out for one another. :confused: Why not make them consistent?

smarlowucf:
Updated to swap the name for ec2 upload.

mash/services/jobcreator/azure_job.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         provider_accounts, provider_groups, requesting_user, last_service,
         utctime, image, cloud_image_name, old_cloud_image_name,
         image_description, distro, download_url, tests, conditions=None,
-        instance_type=None
+        instance_type=None, architecture=None
```
schaefi: How can the architecture be ever a None value ?

schaefi: It should be either mandatory or take a default architecture name but not None

mash/services/jobcreator/schema.py
```diff
@@ -246,6 +246,9 @@
             'type': 'array',
             'items': {'$ref': '#/definitions/non_empty_string'},
             'minItems': 1
+        },
+        'architecture': {
+            'enum': ['x86_64', 'aarch64']
```
schaefi: As this is the architecture used for the cloud it should be 'arm64' for EC2. Unfortunately they don't use consistent names compared to the kernel architecture. Therefore I also suggest a mapping table in the Defaults area to allow a proper translation

mash/services/uploader/cloud/amazon.py
```diff
@@ -63,12 +63,16 @@ class UploadAmazon(UploadBase):
     """
     def post_init(self):
         self.region = 'eu-central-1'
+
+        if self.arch == 'aarch64':
+            self.arch = 'arm64'
```
schaefi: I don't think we should do this here. It would be better to have a translation table in the Defaults space and apply that here. In the schema of the job description we should use the names as they are set by the cloud providers. The way it's done here is vice versa and that's imho confusing

schaefi: You might also consider to make the arch name clear in the attribute in the schema. Meaning I would use:

* cloud_architecture

instead of architecture, to make it crystal clear this is not necessarily the same name like the kernel interprets the processor architecture

rjschwei:
> > We need "aarch64" for looking up the image name, but for image creation in EC2 it needs to be "arm64". I don't see how we get around the mapping. If we allow "arm64" as a value for arch then that will not work with finding the image from the build service.
> 
> I figured they could be swapped out for one another.  Why not make them consistent or is it an EC2 requirement for arm64 rather than ec2utils?

The AWS team that implemented this decided that the name should be "arm64", we were not consulted ;) , and yes it is against everything that everyone else names the beast.

mash/services/uploader/cloud/amazon.py
rjschwei: I agree good point

mash/services/jobcreator/azure_job.py
rjschwei: The default is set elsewhere, but from a code reading perspective I agree.

mash/services/jobcreator/schema.py
rjschwei: Unless the Default implementation knows the framework I think Defaults i snot the proper place for the architecture name transformation, I think it belongs in framework specific code, i.e. the if condition should exist in the uploader in the EC2 code branch.

smarlowucf: If the naming difference is an Amazon choice than I agree it should be transformed in Amazon code rather than defaults.

smarlowucf:
Updated arg name and set default arch in all classes.

smarlowucf:
Previously missed GCE job. Added architecture option to class.

smarlowucf:
Re-based to fix conflicts.

smarlowucf:
@schaefi Made the changes except lest the conversion in Amazon class. Thoughts?

PR #379, opened 2019-01-25
Add cleanup_images option.

### LINKED ISSUE #376: deletion should be optional on azure replication jobs
For some Azure replication jobs, it is required that the original ARM upload be left in place, even after copying to ASM storage for publishing.

`AzureReplicationJob` should accept an argument defining whether or not the 'cleanup' step occurs, and this argument should be exposed as a defined parameter in the job docs.

smarlowucf:
Pass option to replication jobs, Azure replication copies images from ARM to ASM and the ARM image artifacts are removed by default. They can be preserved by providing `cleanup_images==False`.

Resolves #376

### DIFF
--- a/mash/services/jobcreator/azure_job.py
+++ b/mash/services/jobcreator/azure_job.py
@@ -17,6 +17,7 @@
 #
 
 from mash.services.jobcreator.base_job import BaseJob
+from mash.utils.json_format import JsonFormat
 
 
 class AzureJob(BaseJob):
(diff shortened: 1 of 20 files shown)


smarlowucf:
> But wondering if the publisher couldn't at least log a message or even error out if we make that mistake

I think that's very tailored to our use case rather than a general ASM publish job. While we are the first users of MASH I don't think an optional workflow should be forced in a given direction in code based on job type.

smarlowucf:
Re-based to resolve conflicts.

smarlowucf:
@rjschwei Added changes to cleanup unused info in messages. Merging so this can all be re-based again into the open pull requests as everything is conflicting. Reference: https://github.com/SUSE-Enceladus/mash/pull/379/commits/940f135c915837245927ae0fda88ab140fcf6925.

PR #380, opened 2019-01-25
Wait for obs conditions to be met.

### LINKED ISSUE #378: OBS conditions should cause a wait rather than auto fail
If conditions are supplied for a job and they are not met then the OBS service should wait until they become met to continue with image.

The current implementation fails eagerly if the conditions aren't met and fails the job if it's a one time job.

smarlowucf:
Rather than fail eagerly. Log only once that obs service is waiting on conditions for given job.

Fixes #378

### DIFF
--- a/mash/services/obs/build_result.py
+++ b/mash/services/obs/build_result.py
@@ -291,38 +291,48 @@ class OBSImageBuildResult(object):
     def _update_image_status(self):
         try:
             self.iteration_count += 1
+            conditions_fail_logged = False
             self._log_callback('Job running')
-            packages = self._lookup_image_packages_metadata()
-            for condition in self.image_status['conditions']:
-                if 'image' in condition:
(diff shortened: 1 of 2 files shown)


mash/services/obs/build_result.py
```diff
@@ -291,38 +291,48 @@ def _log_error(self, message):
...
-                self.image_status['job_status'] = 'failed'
+                    self.image_status['packages_checksum'] = packages_checksum
+                    self.image_status['job_status'] = 'success'
+                    break
+                else:
+                    if not conditions_fail_logged:
+                        self._log_callback('Waiting for conditions to be met.')
+                        conditions_fail_logged = True
+
+                    time.sleep(120)
```
rjschwei: Image build take more than 2 minutes you can bump this to 5 minutes.

smarlowucf:
Pending comments in #378

PR #383, opened 2019-01-30
Implement GCE deprecation class.

smarlowucf:


mash/services/deprecation/gce_job.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Deprecate image in all target regions in each source region.
+        """
+        self.status = SUCCESS
+
+        if self.old_cloud_image_name is None:
+            # There is no old image that needs deprecation for the job.
+            return
+
+        for account in self.deprecation_regions:
```
rjschwei: This is confusing, maybe a comment is warranted. In GCE images are a global resource and therefore deprecation_regions would always be a collection of 1. While looping over a collection of 1 is of course valid, given the knowledge that images are global it still makes me wonder when reading the code why we would want the loop construct.

mash/services/deprecation/gce_job.py
```diff
@@ -0,0 +1,92 @@
...
+    ):
+        super(GCEDeprecationJob, self).__init__(
+            id, deprecation_regions, last_service, provider, utctime,
+            old_cloud_image_name=old_cloud_image_name, job_file=job_file
+        )
+        self.months_to_deletion = months_to_deletion
+
+    def _deprecate(self):
+        """
+        Deprecate image in all target regions in each source region.
```
rjschwei: Image handling is a global property in GCE, thus speaking of deprecation per region is confusing.

mash/services/jobcreator/gce_job.py
```diff
@@ -88,12 +90,35 @@ def _get_account_info(self):
...
     def get_deprecation_regions(self):
         """
         Return list of deprecation region info.
-
         """
-        raise NotImplementedError('TODO')
+        deprecation_regions = []
+
+        for source_region, value in self.target_account_info.items():
+            deprecation_regions.append(value['account'])
```
rjschwei: This is also confusing, we are mapping account information, i.e. "value['account']" into something that is named _regions where I would expect something named with _regions to hold values such as "us-east-1". I think the naming needs changing.

smarlowucf: This is normal throughout MASH. Each job can be tied to multiple accounts and each account is tied to a source region. This region is where the image is uploaded, tested, replicated from etc etc.

As a note I originally named all of these similar to deprecation_accounts and in a review they were changed to deprecation_regions. So this discussion has been hashed but I think either way it's a confusing name.

smarlowucf: If we only rename for GCE then this looks like:

```
deprecation_accounts = []

         for source_region, value in self.target_account_info.items():
            deprecation_accounts.append(value['account'])
```

and the structure of the target_account_info is provided in the class:

```
Example: {
            'us-west1': {
                'account': 'acnt1',
                'bucket': 'images',
                'family': 'sles-15'
            }
        }
```

mash/services/deprecation/gce_job.py
smarlowucf: Sorry copy paste issue from ec2.

mash/services/deprecation/gce_job.py
rjschwei: Well one option is to use

`account = self.deprecation_regions[0]
# In GCE images are global and we deal with one account`

That would eliminate the "why the loop" question.

rjschwei: Great I guess even in a code block a hashtag makes things bold :P

smarlowucf:
Re-based to include provider -> cloud change.

PR #391, opened 2019-02-08
Move start_jobs to base pipeline class

smarlowucf:
- Rename job methods to process_job so start_job in service can move to base class.
- Rename base job classes to {service}_job.py to be verbose.


smarlowucf:
PipelineService assumes you are using a class for jobs or at least something that acts like a class. I think uploader could move over if the jobs move to job classes. OBS doesn't necessarily fit into this directly because it has no listener messages. I'm still pondering how best to organize things including credentials service and logger.

> The part I don't understand is why the public API of PipelineService uses private methods (e.g _add_job) to expose those methods the subclass has to implement.

Agree this is wrong in naming convention something I will mark to clean up.

PR #393, opened 2019-02-08
Ack job doc message only after the job is saved.

smarlowucf:
Once the job is persisted to disk it's okay to remove the message from queue by responding with acknowledgement. This maintains fault tolerance if system goes down whilst processing the msg.


schaefi:
I'll merge this since I'm testing obs and uploader in my podman container instance. Hope that's ok for you

smarlowucf:
> Hope that's ok for you

Yep, thanks.

PR #394, opened 2019-02-08
Move listener message args to list.

smarlowucf:
And remove the getter method. Default arg is cloud_image_name others can be appended by child classes.


smarlowucf:
Merged with master to fix conflicts.

PR #411, opened 2019-02-25
Add Pull Request Template

domvieira:

This PR adds support for a github pull request template for Mash. The goal is to help standardize changes that go through the PR process and to help provide documentation for those changes. This is the first rev of the template and should go through iteration as it makes sense.

Tested locally, formatting copied from other repos. Will test manually upon merge.

Does not require deployment. Development change only.

### Gif (optional)

Include an optional gif for good luck!

A gif for good luck

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,10 @@
+### What does this PR do? Why are we making this change?
+
+
+### How will these changes be tested?
+
+
+### How will this change be deployed? Any special considerations?
+
+
+### Additional Information


domvieira:
Thanks for the feedback @smarlowucf. I like the question format you chose. I'll adjust my template to match that style and request a re-review.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+### How will this change be deployed? Any special considerations?
+
+
+### Additional Information
+
+
+
+### Gif (optional)
+
+![A gif for good luck](https://media.giphy.com/media/25nKGzahWBeLe/giphy.gif)
```
rjschwei: Please drop the gif

rjschwei:
The PR has to be signed, https://suse-enceladus.github.io/git/ so we probably should include this link in the template as well

domvieira:
Thanks @rjschwei. I'll update the template to remove the gif. One thing we might want to consider, if we're not doing it already, is configure protected branches and require commit signing. That will prevent push if it's not signed. https://help.github.com/en/articles/about-required-commit-signing 

We can talk about that separately though. I'll ask for re-review once the changes are done.

rjschwei:
The master branch is set up to only allow signed commits. Since we have no version branches there's nothing else to do, I think.

PR #413, opened 2019-02-28
Add testing account option

smarlowucf:
Simplify user account handling. Instead of credentials sending the account info for a cloud it parses the specific accounts for a job and passes only that data. This means job creator classes do not duplicate account parsing functionality.

Add testing account option to all cloud accounts. A testing account option can be added when creating a new account to be used for launching instances if it's different than the publishing account.

Unit tests on Travis and integration tests locally.

Any existing accounts that need updated with testing account option must be updated with CLI or manually.

### DIFF
--- a/mash/services/api/schema.py
+++ b/mash/services/api/schema.py
@@ -76,7 +76,8 @@ add_account_azure = {
         },
         'destination_storage_account': {
             '$ref': '#/definitions/non_empty_string'
-        }
+        },
+        'testing_account': {'$ref': '#/definitions/non_empty_string'},
     },
     'additionalProperties': False,
     'required': [
@@ -121,6 +122,7 @@ add_account_ec2 = {
         'partition': {'$ref': '#/definitions/non_empty_string'},
         'region': {'$ref': '#/definitions/non_empty_string'},
         'cloud': {'enum': ['ec2']},
+        'testing_account': {'$ref': '#/definitions/non_empty_string'},
         'requesting_user': {'$ref': '#/definitions/non_empty_string'},
     },
     'additionalProperties': False,
@@ -165,6 +167,7 @@ add_account_gce = {
         },
         'group': {'$ref': '#/definitions/non_empty_string'},
         'cloud': {'enum': ['gce']},
+        'testing_account': {'$ref': '#/definitions/non_empty_string'},
         'region': {'$ref': '#/definitions/non_empty_string'},
         'requesting_user': {'$ref': '#/definitions/non_empty_string'}
     },
--- a/mash/services/api/schema.py
+++ b/mash/services/api/schema.py
@@ -76,8 +76,7 @@ add_account_azure = {
         },
         'destination_storage_account': {
             '$ref': '#/definitions/non_empty_string'
-        },
-        'testing_account': {'$ref': '#/definitions/non_empty_string'},
+        }
     },
     'additionalProperties': False,
     'required': [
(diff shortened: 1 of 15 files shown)


rjschwei:
Hmm, not really what I had in mind, but I understand why it has to be this way, i.e. we have to link the name of the testing account to the "image target account" at the account level since we have partitions in EC2 and one testing account will simply not do the trick. But since we have to have the link between "image target account" and "testing account" at the account level I think in the jobdoc implementation the use of the testing account simply becomes a flag, i.e.

```
  'use_testing_account': True/False
```

with False being the default. If this is not used as a flag then the question would arise, what does it mean to specify `"testing_account": "foo"` when account `bar` is associated with a specific image account and the account `foo` also exists. Logically we should obey the user's specification, i.e. use the test account `foo` over the, at the account level, linked test account but that behavior could no universally apply given the way things are in AWS.

I think we have a couple of options.
1.) Do not link the "test account" and the "image target account" at the account level. This implies that the `testing_account` implementation gets a similar structure to the `cloud_account` setup and it is the users responsibility that the order matches, i.e.

```
  "cloud_accounts" : [
    {  "name": "imagepub" ......},
    { "name": "govpub"....}
  ]
.....
  "testing_accounts" : [
    { "name": "testimagepub"....},
    { "name": "govtest"....}
  ]
```
Now the user has to order the accounts properly, arguably a disadvantage. However, the advantage is that we do not have to link the "testing account" and the "image target account" at the back end in the credentials service, as in the current implementation.

2.) Keep the link at the accounts level as it is, but then turn `testing_account` into a flag and rename it to `use_testing_account` If we don't have the flag then we get into the question pointed out earlier.

I am leaning toward option 1 as it will allow us to handle the "image target account" and the "testing account" completely independently.

smarlowucf:
This is generally confusing now. I think we need to take a step back and first decide on what use cases are needed then look at what functionality needs to be added/modified.

1) GCE has publisher accounts which means images are publicly available as soon as they are created. These accounts have no way to run VMs for testing. To test images we need a testing account.

2) Azure and Amazon have a publish process aside from image creation. An account can be used for testing and publishing.

The goal here is to solve the issue with GCE testing.

The solution for GCE I don't think fits the other cloud frameworks. Not even as an option because in those you have to publish an image to make it available to another account right?

If I create an image in one ec2 account I cannot test it in another account until the image has been shared. And to share the image you publish it visibly to that account right? If so we open a large can of worms going down this road. Now we have two publish steps in one job.

I think we are over-complicating things. Or I'm just confused.

smarlowucf:
Given the complexity I suggest this PR changes to just allow test accounts for GCE. And if we need the functionality for other cloud frameworks then a new discussion can be had.

>> 1.) Do not link the "test account" and the "image target account" at the account level.

Don't like this option. If you think about this information at a data model level it makes sense that the testing account is an account level decision and it doesn't really change on a per job basis. When I create a publishing account for GCE in MASH I know there needs to be a testing account.

Also it's very poor user design to have two lists that implicitly need to matchup in some. Aside from the user experience I've seen issues in python 3.4 and 3.5 where the order of a list of dictionaries hasn't been guaranteed.

>> 2.) Keep the link at the accounts level as it is, but then turn testing_account into a flag and rename it to use_testing_account

This is where I think things are becoming over-complicated by trying to solve this "problem" for ec2 and azure even though it only exists in GCE. In GCE with publishing accounts a testing account is needed. If it's set then it needs to be used.

smarlowucf:
Remove testing account from ec2 and azure. Only fits with GCE workflow.

PR #425, opened 2019-03-15
Azure publishing and deprecation

smarlowucf:

- Rename version_key to vm_images_key to prevent confusion
- Add publish_offer option for azure jobs. By default the publish step will not be triggered. This step
publishes the entire offer which will break other jobs that may be updating different skus at the same time.
- Integrate deprecation for azure. If the old image name matches a version in sku set it to not show
in gui.

Unit & integration tests.

### DIFF
--- a/mash/services/azure_utils.py
+++ b/mash/services/azure_utils.py
@@ -430,7 +430,7 @@ def request_cloud_partner_offer_doc(credentials, offer_id, publisher_id):
 
 def update_cloud_partner_offer_doc(
     doc, blob_url, description, image_name, label, sku,
-    version_key='microsoft-azure-corevm.vmImagesPublicAzure'
+    vm_images_key='microsoft-azure-corevm.vmImagesPublicAzure'
 ):
     """
     Update the cloud partner offer doc with a new version of the given sku.
@@ -456,10 +456,10 @@ def update_cloud_partner_offer_doc(
         if doc_sku['planId'] == sku:
             release = release_date.strftime("%Y.%m.%d")
 
-            if version_key not in doc_sku:
-                doc_sku[version_key] = {}
+            if vm_images_key not in doc_sku:
+                doc_sku[vm_images_key] = {}
 
-            doc_sku[version_key][release] = version
+            doc_sku[vm_images_key][release] = version
             break
 
     return doc
--- a/mash/services/azure_utils.py
+++ b/mash/services/azure_utils.py
@@ -465,6 +465,34 @@ def update_cloud_partner_offer_doc(
     return doc
 
 
+def deprecate_image_in_offer_doc(
+    doc, image_name, sku,
+    vm_images_key='microsoft-azure-corevm.vmImagesPublicAzure'
+):
+    """
+    Deprecate the image in the cloud partner offer doc.
+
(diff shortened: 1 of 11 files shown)


mash/services/azure_utils.py
```diff
@@ -465,6 +465,34 @@ def update_cloud_partner_offer_doc(
...
+):
+    """
+    Deprecate the image in the cloud partner offer doc.
+
+    The image is set to not show in gui.
+    """
+    matches = re.findall(r'\d{8}', image_name)
+
+    if matches:
+        release_date = datetime.strptime(matches[0], '%Y%m%d').date()
```
bear454: In the past, the release_date hasn't always matched the image version, due to the time it takes to publish, vs. when the images were downloaded from the build service.

test/unit/services/deprecation/azure_job_test.py
```diff
@@ -1,16 +1,121 @@
...
+
+        mock_request_doc.return_value = {
+            'definition': {
+                'plans': [
+                    {
+                        'planId': '123',
+                        self.job.vm_images_key: {
+                            '2018.09.09': {
+                                'label': 'New Image 20180909'
+                            }
```
bear454: The `media_name` attribute is an _exact match_ for the `old_cloud_image_name` attribute in the jobdoc. Consider at least validating against this attribute.

bear454: For example:

```
                "microsoft-azure-corevm.vmImagesPublicAzure": {
                    "2018.08.22": {
                        "description": "Pay-as-you-go image. \n\nNew and experienced Linux users get the most usable Linux distribution and stabilized operating system with openSUSE's regular release. Receive updates and harden your OS with openSUSE's latest major distribution. Enjoy the platform of choice for Linux developers, administrators and software vendors.\n\nCustomization of this virtual machine (VM) image can be done using the open source code in openSUSE Build Service.",
                        "label": "openSUSE Leap 42.3",
                        "lunVhdDetails": [],
                        "mediaName": "suse-opensuse-leap-42-3-v20180822",
                        "osVhdUrl": "https://susesuselinuxenterprises.blob.core.windows.net/vm-images/suse-opensuse-leap-42-3-v20180822.vhd?st=2018-11-20T00%3A00%3A00Z&se=2018-12-20T00%3A00%3A00Z&sp=rl&sv=2018-03-28&sr=c&sig=xJpSDrPT/Re3IHzRTmTlYJWKWLqWHmmw1TLzxuHHIlI%3D",
                        "publishedDate": "8/22/2018",
                        "showInGui": true
                    }
                },
```

smarlowucf: If they don't match what should be the course of action? Leave as is with a warning log, raise exception and fail or just do nothing and continue?

bear454: If the `media_name` in CPP doesn't match the `old_cloud_image_name`, it should not deprecate the  image (as it would be deprecating _some other image_)... IMO it should warn and continue.

smarlowucf: :+1:

PR #427, opened 2019-03-19
Use full extension for ec2 and azure image files.

smarlowucf:

Fix image retrieval for ec2 and azure. Use full extension for xz (vhdfixed, raw) image files.

Unit & integration tests.


smarlowucf:
@rjschwei Did some more testing here and moved the prefix check. This still ensures the build number is correct but simplifies the extension check.

smarlowucf:
> Please follow up with any relevant changes to docs & samples.

There's no changes to functionality as it was. This is just to fix a bug that was introduced.

PR #430, opened 2019-03-28
Add more details to readme.

smarlowucf:

- Add more details to readme.
- Move contributing to a separate doc.
- Add more info to contributing doc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,116 @@
+## Dev Requirements
+
+- pytest
+- pytest-cov
+- coverage
+- flake8
+- bumpversion
+- sphinx
+- sphinx_rtd_theme
+- sphinxcontrib-spelling
+- pyenchant
+- travis-sphinx
+- ghp-import
+
+## Contribution Checklist
+
+- All patches must be signed. [Signing Commits](#signing-commits)
+- All contributed code must conform to flake8. [Code Style](#code-style)
+- All new code contributions must be accompanied by a test.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,116 @@
...
+- pyenchant
+- travis-sphinx
+- ghp-import
+
+## Contribution Checklist
+
+- All patches must be signed. [Signing Commits](#signing-commits)
+- All contributed code must conform to flake8. [Code Style](#code-style)
+- All new code contributions must be accompanied by a test.
+    - Tests must pass and coverage remain above 100%. [Unit & Integration Tests](#unit-&-integration-tests)
```
rjschwei: Hmm, "above 100%" makes really no sense. I presume that a code path that is tested multiple times pushes the coverage to a number greater than 100% but that is simply incorrect. I think this should be re-worded, maybe "Test coverage needs to remain at 100%"

CONTRIBUTING.md
```diff
@@ -0,0 +1,116 @@
...
+$ git tag -a v{version}
+# git tag -a v0.0.1
+
+# Create a message with the changes since last release and push tags.
+$ git push --tags
+```
+
+## Unit & Integration Tests
+
+All tests should pass and test coverage should remain above 100%.
```
rjschwei: dito

CONTRIBUTING.md
smarlowucf: Copy and paste mistake.

smarlowucf:
@rjschwei Good?

PR #435, opened 2019-04-09
Move account/credentials functions to new datastore class

smarlowucf:

I realize this is a very big change based on LOC but the vast majority is moving methods as is from one class to another. There are very few functional changes in the code that's moved except the function signatures changed where necessary.

Moves account handling functions from the credentials service to a new datastore class. The datastore class wraps all information about the datastore and credentials storage which includes encryption and key rotation.

The new credentials service contains only knowledge of public interfaces to get/set/retrieve account/credentials information. It injects cloud specific account info when adding an account using a factory style method.

The account class structure was completely removed as it was overly complicated. There was no need for any shared methods. Also, a few other legacy modules which were never used are removed.

Finally, the utility methods for key rotation where also moved directly from the key rotation module to the new datastore class.

Unit testing + integration testing locally: Tested account add/delete with/without groups for all three clouds + Valid & invalid job submission.


mash/services/credentials/__init__.py
```diff
@@ -15,28 +15,43 @@
...
+    elif cloud == CSP.gce:
+        account_info = {
+            'bucket': message['bucket'],
+            'region': message['region'],
+            'testing_account': message.get('testing_account')
+        }
+    else:
+        raise MashCredentialsException(
+            'CSP {0} is not supported.'.format(cloud)
+        )
```
rjschwei: The account_info should be in a dict to avoid the long if-elif block:

```
account_data = {
    CSP.ec2 = ('additional_regions', 'partition', 'region'),
    CSP.azure = (....),
    .....
}

account_msg_keys = account_data.get(cloud)
if not account_msg_keys :
    raise ....
```

The use comprehension to form the account_info dict from the keys that are in the tuple.

PR #441, opened 2019-04-11
Remove unused uploader conventions classes.

smarlowucf:

Remove unused uploader conventions classes.

Unit tests.


smarlowucf:
> The idea with this was to force us on our own naming conventions, especially because we apply that when searching the image on obs by the given name in the job document.

I like the idea but it wasn't used yet. If it ends up a priority in the future we could add this back and maybe put in base folder to be used by all services.

PR #443, opened 2019-04-12
Create service classes as Pipeline instances

smarlowucf:

This is the first step for refactoring the pipeline services. Since the services themselves share the same functions aside from some config changes, these differences are now injected at service class creation. This removes the need for a service class for each service. The jobs are created from a single class factory based on cloud and service exchange.

Unit tests + integration testing locally with successful/failed jobs through deprecation.


mash/services/__init__.py
```diff
@@ -56,3 +93,44 @@ def get_configuration(service, config_file=None):
         raise MashConfigException(
             'No configuration available for {0} service.'.format(service)
         )
+
+
+class Job(object):
```
rjschwei: Since we speak of a "job" as an entity that the pipeline processes and that is submitted via a "job document" choosing the "Job" name for the factory implementation is confusing. Also something we used to call a "service" is now being replaced in nomenclature with "Job" so "deprecation_service" no becomes "AzureDeprecationJob"

This should have been discussed in some way shape or form, this is a very large change.

Last but not least, given the hiccups we experienced during first use of mash I am not certain that now is a good time to undertake such large changes. We've not had the opportunity to test the changes that were made based on our first experience with mash.

This should be put on hold until we have a chance to discuss, and have additional experience wit hthe current state of mash.

Stopping my review here.

smarlowucf: > Since we speak of a "job" as an entity that the pipeline processes and that is submitted via a "job document" choosing the "Job" name for the factory implementation is confusing. Also something we used to call a "service" is now being replaced in nomenclature with "Job" so "deprecation_service" no becomes "AzureDeprecationJob"

The new Job factory is for creating an instance of a job and it's used by the pipeline service implementation. This was to allow for simplification of service implementation. E.g. Since the only difference between TestingService and ReplicationService for example was job creation now there's a single method to "inject" that class creation. This means instead of needing a bunch of empty classes that extend from PipelineService now an instance of TestingService is created like `PipelineService(service_exchange='testing')`.

This PR is big, but it's broken out into smaller commits that build on each other. Reviewing each one-by-one I think will give a better idea of what's changing.

> This should have been discussed in some way shape or form, this is a very large change.

This was discussed albeit briefly in a chat with Marcus last week.

> Last but not least, given the hiccups we experienced during first use of mash I am not certain that now is a good time to undertake such large changes. We've not had the opportunity to test the changes that were made based on our first experience with mash.

Yes there have been setbacks in MASH in many areas. However, I think it's bad to say we should stop making changes and improvements to the code base until things are tested again. If we are weary to make changes because of concern for breakage then this is an issue in CI/CD and not in the speed of development. Of course I test the changes but not every single use case for every PR. Maybe this is a point to work on/discuss a CI/CD pipeline for automated testing of the MASH code base itself.

> This should be put on hold until we have a chance to discuss, and have additional experience wit hthe current state of mash.

The end goal with these changes is to get the pipeline service workflow into a satisfactory state which then gives me the confidence to migrate uploader service. The workflow of the pipeline service itself has been very stable since we started hammering MASH with testing. The issues have mainly been due to missing/mis-communicated business rules, job class implementation bugs and account handling bugs.

rjschwei: > This was discussed albeit briefly in a chat with Marcus last week.

How does that help me as a reviewer when the ideas were not communicated via meeting minutes or other means?

> I think it's bad to say we should stop making changes and improvements to the code base until things are tested again.

I did not say we should stop, I shared my thoughts about being uncertain whether this is a good idea at this time.

smarlowucf: > How does that help me as a reviewer when the ideas were not communicated via meeting minutes or other means?

Fair, this was prior to the meeting so not in the minutes.

> I did not say we should stop, I shared my thoughts about being uncertain whether this is a good idea at this time.

I think there's a difference when it comes to code cleanup and refactor where the functionality is not changing. To me these are all "private" code changes. They shouldn't affect any public functionality or interface. So what level of oversight do we need for these type of changes?

mash/services/__init__.py
```diff
@@ -56,3 +93,44 @@ def get_configuration(service, config_file=None):
...
+    """
+    Service Job Factory.
+
+    Attributes:
+
+    * :attr:`cloud`
+        cloud service provider name
```
rjschwei: s/service provider//

mash/services/__init__.py
```diff
@@ -56,3 +93,44 @@ def get_configuration(service, config_file=None):
...
+            )
+
+        try:
+            job = job_class(job_config, service_config)
+        except Exception as error:
+            raise MashJobException(
+                'Invalid job configuration: {0}'.format(error)
+            )
+
+        return job
```
rjschwei: I think the Job class, which is really the factory should not live in services/__init__.py it should live in it's own implementation file properly named job_factory.py. The implementation should then have a  method to create the job. This follows commonly used factory design patterns. Generally although possible as shown here, the factory is not implemented in the constructor of the class. In terms of the client that would look as follows:

```
from service.job_factory import JobFactory

job = JobFactory.create_job(...)
```

For factory pattern implementation in Python this can be used as a reference https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Factory.html in the example the method is named "factory" in other places C++ and Java the methods for generating the proper instances are often called "create*" a naming convention I prefer and used in the example.

Anyway, I think we shold follow commonly accepted practice and not implement the factory in the constructor.

mash/services/pipeline_service.py
```diff
@@ -51,17 +61,46 @@ def post_init(self):
...
+                if 'job_file' not in job_config:
+                    job_config['job_file'] = self.persist_job_config(
+                        job_config
+                    )
+                    job.job_file = job_config['job_file']
+
+                self.log.info(
+                    'Job queued, awaiting listener message.',
+                    extra=job.get_job_id()
+                )
```
rjschwei: Two options here, either optimize for the most common expected case, which should be the none duplicate job-ids or drop the else branch.

either:
```
if not job_id in self.jobs:
  # DO WHAT'S IN THE CURRENT ELSE CLAUSE
else:
  # DO WHAT'S IN THE CURRENT IF CLAUSE
```
or

```
if job_id in self.jobs:
            self.log.warning(
                'Job already queued.',
                extra={'job_id': job_id}
            )
           return
```

The way branching instructions works out "if-else" generate more jumps then an "if return no else" implementations. So the advantage of the second implementation is that there are fewer jumps in the generated code, but the intermediate return can also be hard to catch when reading the code.

From a code reading perspective the most common and expected case should be in the if condition. Having jobs with duplicated IDs should be the exception; and given that having duplicate job-ids should probably throw. Anyway, since the non-duplicate id should be the preferred expected code path it should be in the if clause if we want to stick with the if-else branching structure.

smarlowucf:
Rebased to fix merge conflict and updated PR per comments.

PR #445, opened 2019-04-25
- Add new AWS region ap-east-1

rjschwei:

Add new region for AWS


rjschwei:
Helper image for ap-east-1 added

PR #448, opened 2019-05-03
Use service exchange name for log module name.

smarlowucf:

Use service exchange name for log module name.

Local integration tests.


schaefi:
LGTM

PR #449, opened 2019-05-03
Migrate uploader classes to pipeline workflow

smarlowucf:

The uploader service is now an instance of PipelineService and the job classes extend from mash_job base class. Moved the job class modules to match the other PipelineService instances.

Unit and integration tests. Locally tested successful and failed jobs for all three clouds through testing service.


schaefi:
@smarlowucf thanks for your effort on this

PR #450, opened 2019-05-08
Handle any uncaught exceptions from datastore.

smarlowucf:

Handle any uncaught exceptions from datastore. Catch and log the exception. This prevents the service from crashing and also ensures the messages are ack'ed if there is a failure.


smarlowucf:
Fixed merge conflicts. Good?

PR #452, opened 2019-05-09
GCE family is optional.

smarlowucf:

GCE family is optional.

If it is not provided in job doc do not use it when creating image.

### DIFF
--- a/mash/services/api/schema.py
+++ b/mash/services/api/schema.py
@@ -385,7 +385,6 @@ gce_job_message['properties']['months_to_deletion'] = {
 gce_job_message['properties']['testing_account'] = {
     '$ref': '#/definitions/non_empty_string'
 }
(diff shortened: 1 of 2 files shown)


rjschwei:
Generally correct for mash, but not for our use-case. I guess this was inevitable, I think we need a plugin mechanism for job doc verification.

The SUSE condition would be "if $PROJECT == $RELEASE_PROJECT then --family must be set" I am OK with this change, but I would like a "TODO" for the additional verification plugin and then see this condition implemented.

smarlowucf:
Rebased to fix conflicts. Created https://github.com/SUSE-Enceladus/mash/issues/456 to document comment. This would be an account level flag added during gce cloud account creation. Each GCE cloud account in MASH is a service account which is only for a specific GCE project.

PR #453, opened 2019-05-09
Move credentials methods to pipeline service.

smarlowucf:

Move credentials methods to pipeline service. Credentials methods are not used by all base services.

Unit and integration tests.

### DIFF
--- a/mash/services/pipeline_service.py
+++ b/mash/services/pipeline_service.py
@@ -17,12 +17,16 @@
 #
 
 import json
+import jwt
 
 from amqpstorm import AMQPError
 
 from apscheduler import events
 from apscheduler.jobstores.base import ConflictingIdError, JobLookupError
 from apscheduler.schedulers.background import BackgroundScheduler
+
+from cryptography.fernet import Fernet, MultiFernet
+from datetime import datetime, timedelta
 
 from pytz import utc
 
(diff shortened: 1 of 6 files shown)


smarlowucf:
Re-based to fix merge conflicts.

mash/services/pipeline_service.py
```diff
@@ -421,6 +436,128 @@ def _process_msg_arg(self, listener_msg, arg, job):
...
+            self.log.error(
+                'Invalid credentials response token: {0}'.format(error)
+            )
+        else:
+            return job_id, decrypted_credentials
+
+        # If exception occurs decoding credentials return None.
+        return None, None
+
+    def decrypt_credentials(self, credentials):
```
rjschwei: The client should not be responsible for decrypting the credentials. The only service that should know anything about encryption is the credentials service as it is integral to the credentials service function and implementation. The credential service stores the user credential info on disk and thus they need to be encrypted at that point.

I realize this just moves the existing implementation, but I am thinking we didn't get the separation of concerns right previously.

The credentials are in a JWT token, thus "protected" from prying eyes. An attacker that has access to the message bus will most likely also have access to the storage and thus the encryption key. Therefore the separation of concerns, i.e. credentials service does encrypt/decrypt vs. everyone does decrypt should have a higher weight.

I think the flow in the credential service should be decrypt, create JWT, put the JWT on the bus.

smarlowucf: This PR is simply moving methods around so any functional changes should follow in a new PR. On that note with the addition of authentication the credentials and user management will need to move thus any changes should probably coincide with that move. Also, JWT tokens are only signed they are not encrypted so the data is not protected.

rjschwei: Correct about JWT not being encrypted, the point was if someone can sniff the message bus they have access to the system, at least in our current assumptions about he mash implementation where we do not share the message bus with other applications. So if the attacker has access to the message bus the attacker also has access to the encryption key as we store it on the system disk rather than using some service. Therefore having knowledge about encryption spread to every service is probably overkill.

But yes, any of this should not effect this PR as this is just refactoring. Probably a good discussion point later.

smarlowucf: Not sure if this should be an issue, but will create one to document the discussion. Further discussion https://github.com/SUSE-Enceladus/mash/issues/458.

PR #459, opened 2019-05-20
Migrate job specific methods

smarlowucf:

Migrate all methods specific to handling jobs from base mash service class to the pipeline, obs and credentials service classes. Also, move any related attrs.

Unit tests and local integration testing.

### DIFF
--- a/mash/services/credentials/service.py
+++ b/mash/services/credentials/service.py
@@ -33,6 +33,9 @@ class CredentialsService(MashService):
     Implements CredentialsService based on web token technology
     """
     def post_init(self):
+        self.listener_queue = 'listener'
+        self.job_document_key = 'job_document'
+
         self.set_logfile(self.config.get_log_file(self.service_exchange))
 
         self.encryption_keys_file = self.config.get_encryption_keys_file()
@@ -61,6 +64,9 @@ class CredentialsService(MashService):
         )
         self.bind_queue(
             self.service_exchange, self.delete_account_key, self.listener_queue
+        )
+        self.bind_queue(
+            self.service_exchange, self.job_document_key, self.service_queue
         )
         self._bind_credential_request_keys()
 
--- a/mash/services/credentials/service.py
+++ b/mash/services/credentials/service.py
@@ -34,8 +34,9 @@ class CredentialsService(MashService):
     """
     def post_init(self):
         self.listener_queue = 'listener'
-        self.job_document_key = 'job_document'
         self.service_queue = 'service'
+        self.job_document_key = 'job_document'
+        self.listener_msg_key = 'listener_msg'
 
         self.set_logfile(self.config.get_log_file(self.service_exchange))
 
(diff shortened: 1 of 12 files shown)


mash/services/credentials/service.py
```diff
@@ -373,13 +389,52 @@ def delete_account(self, message):
             requesting_user, account_name, cloud
         )
 
+    def remove_file(self, config_file):
+        """
+        Remove file from disk if it exists.
+        """
+        try:
+            os.remove(config_file)
+        except Exception:
```
rjschwei: I think the specific exception should be caught here, i.e. `except FileNotFoundError:`

mash/services/obs/service.py
```diff
@@ -237,3 +260,29 @@ def _start_job(self, job):
             'ok': True,
             'message': 'Job started'
         }
+
+    def persist_job_config(self, config):
```
rjschwei: Maybe worth to push this into a utility function to be re-used. This is the second implementation of the "persist" code. Not a "must change" item

mash/services/pipeline_service.py
```diff
@@ -558,14 +578,68 @@ def publish_credentials_request(self, job_id):
...
+    def remove_file(self, config_file):
+        """
+        Remove file from disk if it exists.
+        """
+        try:
+            os.remove(config_file)
+        except Exception:
+            pass
+
+    def persist_job_config(self, config):
```
rjschwei: Number 3

mash/services/credentials/service.py
smarlowucf: Fair, actually I thought we had a context manager for ignoring things that are just set to `pass`. Although, we still want to catch general exceptions and maybe those get logged. We don't want an un-handled exception crashing the service.

rjschwei: Well we really only want to ignore the "file not there" problem. If there is another exception that could point to a larger issues. The other exception would be that we cannot remove the file which might imply that the permissions have changed from under us and in that case we probably shouldn't continue as the system would be in a wacky state.

mash/services/obs/service.py
smarlowucf: There is some duplication with these changes. The question is, is the duplication or being tied to a shared util module a bigger issue? As we decided that certain services in MASH would eventually be independently scalable we have to consider where it makes sense to use shared utility methods. An method that is shared by different services (not packaged together) will need to be it's own package which adds complexity.

E.g. if had two packages one for core mash (w/ credentials) and one for pipeline then we need a separate utility package that would hold the persist_job_config that is required by both.

smarlowucf:
Rebased to fix conflicts.

mash/services/obs/service.py
rjschwei: Sorry for the delayed answer. Yes, dependency management comes with it's own set of problems and yes it does add a certain level of complexity. However, the dependency resolution is a well understood and generally resolved problem and therefore I do consider this as only a marginal increase in complexity.

W.r.t. code duplication I can only speak from experience in that code that  is "reused" via duplication or cut and paste in inevitably drifts apart and that is generally a much bigger problem then the added complexity of managing dependencies for sharing code.

Thanks for creating the utility functions.

If we get to the point where mash needs to separate into different packages, Python and or RPM, then we'll deal with the "utils" packaging problem.

smarlowucf: Ack :+1:

PR #464, opened 2019-05-21
Remove references to pint service.

### LINKED ISSUE #457: Remove pint service from MASH
For now pint is not an integrated service in the MASH pipeline. It should be removed from the default list.

https://github.com/SUSE-Enceladus/mash/blob/master/mash/services/base_defaults.py#L45

smarlowucf:

Remove references to pint service. This is not integrated and still is TBD.

Unit tests.

Fixes #457


smarlowucf:
There are already a few line items in Trello for this.

PR #466, opened 2019-05-21
Add is_publishing_account option to gce jobs.

### LINKED ISSUE #456: GCE Family Required for publishing accounts
The GCE family option for a job should be required for accounts that are in publishing projects. A cloud account for GCE is only for a specific project. So when creating a GCE account, MASH should have a flag like `is_publishing_account` to signify this.

When a job is submitted for an account marked as publishing the family attribute should be validated as required.

smarlowucf:

Add is_publishing_account option to gce jobs. If publishing account is used for a job without a family raise exception.

Unit and integration testing.

Fixes #456

### DIFF
--- a/mash/services/jobcreator/gce_job.py
+++ b/mash/services/jobcreator/gce_job.py
@@ -64,6 +64,12 @@ class GCEJob(BaseJob):
                     'Jobs using a GCE publishing account require a family.'
                 )
 
+            if is_publishing_account and not testing_account:
+                raise MashJobCreatorException(
+                    'Jobs using a GCE publishing account require'
+                    ' the use of a testing account.'
+                )
+
             self.target_account_info[region] = {
                 'account': account,
                 'bucket': bucket,
--- a/mash/services/jobcreator/gce_job.py
+++ b/mash/services/jobcreator/gce_job.py
@@ -16,6 +16,7 @@
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
 
+from mash.mash_exceptions import MashJobCreatorException
 from mash.services.jobcreator.base_job import BaseJob
 from mash.utils.json_format import JsonFormat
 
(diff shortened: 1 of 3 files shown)


smarlowucf:
@rjschwei Good?

mash/services/jobcreator/gce_job.py
```diff
@@ -55,9 +56,13 @@ def get_account_info(self):
...
-            testing_account = self.cloud_accounts[account].get(
-                'testing_account'
-            ) or info.get('testing_account')
+            testing_account = info.get('testing_account')
+            is_publishing_account = info.get('is_publishing_account')
+
+            if is_publishing_account and not self.family:
+                raise MashJobCreatorException(
+                    'Jobs using a GCE publishing account require a family.'
+                )
```
rjschwei: Need the same test for testing account:

```
if is_publishing_account and not testing_account:
    raise MashJobCreatorException(
￼	                    'Jobs using a GCE publishing account requires the use of a testing account.'
    )
```

mash/services/jobcreator/gce_job.py
```diff
@@ -55,9 +56,13 @@ def get_account_info(self):
...
 
-            testing_account = self.cloud_accounts[account].get(
-                'testing_account'
-            ) or info.get('testing_account')
+            testing_account = info.get('testing_account')
+            is_publishing_account = info.get('is_publishing_account')
+
+            if is_publishing_account and not self.family:
+                raise MashJobCreatorException(
+                    'Jobs using a GCE publishing account require a family.'
```
rjschwei: s/require/requires/

mash/services/jobcreator/gce_job.py
smarlowucf: This is okay for now since there's not a good spot for account based business rules but will move when auth is integrated and account handling moved.

mash/services/jobcreator/gce_job.py
smarlowucf: Not a typo: Jobs ... require a ... is correct.

PR #467, opened 2019-05-21
Add use root swap option for ec2 upload.

### LINKED ISSUE #465: support rootswap in EC2
Creating on-demand images in EC2 is only possible by either having access to a whitelisted account where one can set the billing code, or one has to use a so called root-swap method.

In an effort to allow the SUSE QA team to use mash for image uploading it is important to allow the creation of on-demand images outside of the whitelisted account space. THerefore the rootswap mechanism needs to be supported by mash.

### LINKED ISSUE #471: root_swap support
The goal of the root_swap support is to allow users to create an on-demand image outside the white listed account that supports billing code registration. This requires that the helper image is an on-demand image. In support of the implementation in #467 we need to support using a different helper AMI when using the root swap method vs the non root swap method. This allows for reduced cost image creation when the creation of on-demand images is not necessary.

That said it is a valid decision for us to say that the cost of using an on-demand image for image creation is negligible and therefore there is no need to have different helper AMIs in the config.

smarlowucf:

Add use root swap option for ec2 upload.

Unit testing.

Fixes #465 and resolves #471

### DIFF
--- a/mash/services/api/schema.py
+++ b/mash/services/api/schema.py
@@ -315,6 +315,7 @@ ec2_job_message['properties']['allow_copy'] = {'type': 'boolean'}
 ec2_job_message['properties']['billing_codes'] = {
     '$ref': '#/definitions/non_empty_string'
 }
+ec2_job_message['properties']['use_root_swap'] = {'type': 'boolean'}
 ec2_job_message['definitions']['account'] = {
     'type': 'object',
     'properties': {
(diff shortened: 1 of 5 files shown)


smarlowucf:
Will add the changes in this PR and request new review...

smarlowucf:
@rjschwei Ready with the ami handling.

PR #469, opened 2019-05-29
Make distro option an enum.

smarlowucf:

IPA validates distro for expected values so MASH will fail early if distro does not match enum (opensuse_leap or sles).


smarlowucf:
> This is fine for now, but is there some way we can enumerate this from `ipa`?

Since MASH requires IPA we have the site package and can import the list from IPA. For now it's pretty static though so I think it's okay to duplicate the values.

PR #478, opened 2019-06-12
Use class based API views

smarlowucf:

Use class based views and add swagger documentation from flask-restplus package. Also, provides better error messages for endpoints.

Unit and integration testing.


smarlowucf:
@rjschwei Cleaned up a few things from discussion and added examples to schema fields.

mash/services/api/routes/accounts/azure.py
```diff
@@ -0,0 +1,79 @@
...
+    account_response,
+    validation_error_response,
+    delete_account_request
+)
+from mash.services.api.schema.accounts.azure import add_account_azure
+from mash.services.api.routes.utils import publish
+
+api = Namespace(
+    'Azure Accounts',
+    description='Azure account related operations'
```
rjschwei: s/related//

mash/services/api/routes/accounts/ec2.py
```diff
@@ -0,0 +1,79 @@
...
+    account_response,
+    validation_error_response,
+    delete_account_request
+)
+from mash.services.api.schema.accounts.ec2 import add_account_ec2
+from mash.services.api.routes.utils import publish
+
+api = Namespace(
+    'EC2 Accounts',
+    description='EC2 account related operations'
```
rjschwei: s/related//

mash/services/api/routes/accounts/gce.py
```diff
@@ -0,0 +1,79 @@
...
+    account_response,
+    validation_error_response,
+    delete_account_request
+)
+from mash.services.api.schema.accounts.gce import add_account_gce
+from mash.services.api.routes.utils import publish
+
+api = Namespace(
+    'GCE Accounts',
+    description='GCE account related operations'
```
rjschwei: s/related//

mash/services/api/routes/jobs/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+from flask import jsonify, make_response
+from flask_restplus import fields, Namespace, Resource
+
+from mash.services.api.schema import validation_error
+from mash.services.api.routes.utils import publish
+
+api = Namespace(
+    'Jobs',
+    description='Job related operations'
```
rjschwei: s/related//

mash/services/api/routes/jobs/azure.py
```diff
@@ -0,0 +1,50 @@
...
+from mash.services.api.routes.jobs import (
+    job_response,
+    validation_error_response,
+    process_job_add_request
+)
+from mash.services.api.schema.jobs.azure import azure_job_message
+
+api = Namespace(
+    'Azure Jobs',
+    description='Azure Job related operations'
```
rjschwei: s/related//

mash/services/api/routes/jobs/azure.py
```diff
@@ -0,0 +1,50 @@
...
+class AzureJobCreate(Resource):
+    @api.doc('add_azure_job')
+    @api.expect(azure_job)
+    @api.response(201, 'Job added', job_response)
+    def post(self):
+        """
+        Add Azure job.
+        """
+        data = json.loads(request.data.decode())
+        data['cloud'] = 'azure'
```
rjschwei: This would make "cloud" in the job doc the user creates unnecessary. This is a fair decision. On the other hand having "cloud" in the data on the user is a nice and easy way for the user to identify by visual inspection for which framework the job is intended for.

I would tend to expect that the user sends the data along and then I think it would be nice to validate that the proper cloud is set rather than to overwrite whatever the user sent. 

Also it seems that the "cloud" value needs to be in the job doc on the user side so the mash cli can determine to which API endpoint to send the request. Otherwise "mash job add" would have to gain a "--cloud" argument.

In short, I don't think we want to overwrite the data we get form the user here, but we want to validate that it is sent to the proper API, i.e.

```
if data['cloud'] != 'azure':
    return 400
```

mash/services/api/routes/jobs/ec2.py
```diff
@@ -0,0 +1,50 @@
...
+from mash.services.api.routes.jobs import (
+    job_response,
+    validation_error_response,
+    process_job_add_request
+)
+from mash.services.api.schema.jobs.ec2 import ec2_job_message
+
+api = Namespace(
+    'EC2 Jobs',
+    description='EC2 Job related operations'
```
rjschwei: s/related//

mash/services/api/routes/jobs/ec2.py
```diff
@@ -0,0 +1,50 @@
...
+class EC2JobCreate(Resource):
+    @api.doc('add_ec2_job')
+    @api.expect(ec2_job)
+    @api.response(201, 'Job added', job_response)
+    def post(self):
+        """
+        Add EC2 job.
+        """
+        data = json.loads(request.data.decode())
+        data['cloud'] = 'ec2'
```
rjschwei: Same as previous.

mash/services/api/routes/jobs/gce.py
```diff
@@ -0,0 +1,50 @@
...
+from mash.services.api.routes.jobs import (
+    job_response,
+    validation_error_response,
+    process_job_add_request
+)
+from mash.services.api.schema.jobs.gce import gce_job_message
+
+api = Namespace(
+    'GCE Jobs',
+    description='GCE Job related operations'
```
rjschwei: s/related//

mash/services/api/routes/jobs/gce.py
```diff
@@ -0,0 +1,50 @@
...
+class GCEJobCreate(Resource):
+    @api.doc('add_gce_job')
+    @api.expect(gce_job)
+    @api.response(201, 'Job added', job_response)
+    def post(self):
+        """
+        Add GCE job.
+        """
+        data = json.loads(request.data.decode())
+        data['cloud'] = 'gce'
```
rjschwei: See previous comments/question

mash/services/api/schema/jobs/ec2.py
```diff
@@ -18,39 +18,38 @@
...
-    'anyOf': [
-        {'enum': ['all', 'none']},
-        account_numbers
-    ]
-}
+ec2_job_message['properties']['share_with'] = share_with
 ec2_job_message['properties']['allow_copy'] = {'type': 'boolean'}
-ec2_job_message['properties']['billing_codes'] = non_empty_string
+ec2_job_message['properties']['billing_codes'] = string_with_example(
+    '1234567890,0987654321'
```
rjschwei: Always starts with "bp-", please include the prefix in the example

mash/services/api/routes/jobs/azure.py
smarlowucf: The endpoints will be split by cloud so having cloud in the job doc is duplicative. It's to the doc prior to sending to the pipeline since all data is in the single json dict.

New endpoints like `mash job add ec2` and `mash job add azure` are needed to provide clean and concise API documenation that describes the correct required/optional args for jobs. These args are different for each cloud framework.

mash/services/api/routes/accounts/azure.py
rjschwei: Really these are "account operations" not "account related operation", but fair enough personal taste.

mash/services/api/routes/jobs/azure.py
rjschwei: Well then there should be an issue for the cli that describes this change so it does not get forgotten.

mash/services/api/schema/jobs/ec2.py
rjschwei: Sorry but that this example start with "bp-" is important just like the example for the root swap image starts with "ami-"

mash/services/api/routes/jobs/azure.py
smarlowucf: Will add one, these changes are only being staged in feature branch until the entire auth change is done. Once it's ready the client and server code can be merged and released together.

smarlowucf:
Removed 'related' from namespaces, updated billing code example and created tracker issue in client for these changes.

PR #488, opened 2019-07-16
Migrate obs service to obs-img-utils

### LINKED ISSUE #407: Progress during image download
We should write some progress indicator to the log for image download. Image downloading can take a while and a user tailing the log has no indication if the download dies or if it is still progressing.

smarlowucf:

OBS download image functions have moved to a separate package. OBS service is updated to use the new package.

Provides better log messages, mutlibuild support and progress during download.

Fixes #407 

Integration and unit tests


.virtualenv.requirements.txt
```diff
@@ -27,3 +27,4 @@ img-proof>=4.0.0
 lxml
 requests
 urllib3<1.25,>=1.20
+obs-img-utils>=0.0.3
```
rjschwei: This should be `<=` based on discussion for obs-img-utils it is likely that the API will change for newer versions

package/mash.spec
```diff
@@ -48,6 +48,7 @@ BuildRequires:  python3-img-proof-tests>=4.0.0
 BuildRequires:  python3-lxml
 BuildRequires:  python3-Flask
 BuildRequires:  python3-requests
+BuildRequires:  python3-obs-img-utils >= 0.0.3
```
rjschwei: dito on the version requirement

Issue #483, opened 2019-06-25
re-run test on temporal failures


rjschwei:
The test may fail due to resource limits, such as

```
RROR 2019-06-25 12:14:16,092 TestingService
    Pass[1]: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/mash/services/testing/img_proof_helper.py", line 95, in img_proof_test
    timeout=img_proof_timeout
  File "/usr/lib/python3.6/site-packages/img_proof/ipa_controller.py", line 132, in test_image
    return cloud.test_image()
  File "/usr/lib/python3.6/site-packages/img_proof/ipa_cloud.py", line 644, in test_image
    self._launch_instance()
  File "/usr/lib/python3.6/site-packages/img_proof/ipa_gce.py", line 220, in _launch_instance
    **kwargs
  File "/usr/lib/python3.6/site-packages/libcloud/compute/drivers/gce.py", line 4017, in create_node
    self.connection.async_request(request, method='POST', data=node_data)
  File "/usr/lib/python3.6/site-packages/libcloud/common/base.py", line 772, in async_request
    response = request(**kwargs)
  File "/usr/lib/python3.6/site-packages/libcloud/compute/drivers/gce.py", line 121, in request
    response = super(GCEConnection, self).request(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/libcloud/common/google.py", line 808, in request
    *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/libcloud/common/base.py", line 637, in request
    response = responseCls(**kwargs)
  File "/usr/lib/python3.6/site-packages/libcloud/common/base.py", line 155, in __init__
    self.object = self.parse_body()
  File "/usr/lib/python3.6/site-packages/libcloud/common/google.py", line 296, in parse_body
    raise GoogleBaseError(message, self.status, code)
libcloud.common.google.GoogleBaseError: {'domain': 'usageLimits', 'reason': 'quotaExceeded', 'message': "Quota 'CPUS' exceeded. Limit: 50.0 in region us-west1."}
```

mash should classify this as a recoverable failure and re-run the tests in some other region.

jgleissner:
AFAICS mash triggers tests in multiple regions, if the job definition has multiple `cloud_accounts` configured. Mash could consider hitting quota restrictions as soft errors and only fail if none of the tests succeed. Makes sense?

smarlowucf:
[...]

PR #489, opened 2019-07-16
testing service: retry test in fallback region

jgleissner:
If img_proof throws IpaRetryableError, and there is at least one fallback region configured, retry the test in a different region (issue  #483).

This adds an optional parameter `fallback_regions` to the GCE job message.

### DIFF
--- a/mash/services/jobcreator/base_job.py
+++ b/mash/services/jobcreator/base_job.py
@@ -52,6 +52,7 @@ class BaseJob(object):
             )
 
         self.tests = kwargs.get('tests', [])
+        self.test_fallback_regions = kwargs.get('test_fallback_regions', [])
         self.conditions = kwargs.get('conditions')
         self.instance_type = kwargs.get('instance_type')
         self.old_cloud_image_name = kwargs.get('old_cloud_image_name')
@@ -207,7 +208,8 @@ class BaseJob(object):
             'testing_job': {
                 'cloud': self.cloud,
                 'tests': self.tests,
-                'test_regions': self.get_testing_regions()
+                'test_regions': self.get_testing_regions(),
+                'test_fallback_regions': self.test_fallback_regions
             }
         }
 
--- a/mash/services/jobcreator/base_job.py
+++ b/mash/services/jobcreator/base_job.py
@@ -53,6 +53,8 @@ class BaseJob(object):
 
         self.tests = kwargs.get('tests', [])
         self.test_fallback_regions = kwargs.get('test_fallback_regions', [])
+        self.test_fallback = 'test_fallback_regions' not in kwargs and not self.test_fallback_regions
+
         self.conditions = kwargs.get('conditions')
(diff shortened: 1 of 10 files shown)


smarlowucf:
This will need to be pinned (>=) to the new IPA version which incorporates the retryable errors before tests can pass:

```
ImportError: cannot import name 'IpaRetryableError' from 'img_proof.ipa_exceptions'
```

mash/services/jobcreator/base_job.py
```diff
@@ -203,7 +204,8 @@ def get_testing_message(self):
             'testing_job': {
                 'cloud': self.cloud,
                 'tests': self.tests,
-                'test_regions': self.get_testing_regions()
+                'test_regions': self.get_testing_regions(),
+                'test_fallback_regions': self.test_fallback_regions
```
smarlowucf: The convention is to only add a key to `testing_job` if it is provided for optional args.

See:

```
if self.distro:
            testing_message['testing_job']['distro'] = self.distro
```

mash/services/testing/gce_job.py
```diff
@@ -50,6 +50,7 @@ def post_init(self):
         """
         try:
             self.test_regions = self.job_config['test_regions']
+            self.test_fallback_regions = self.job_config['test_fallback_regions']
```
smarlowucf: This should also move outside of the try catch block which only contains required args. Similar to distro handling:

```
self.distro = self.job_config.get('distro', 'sles')
```

mash/services/jobcreator/base_job.py
jgleissner: We might as well try all, as the operation isn't really expensive (if it fails due to quota). We could add another parameter `test_max_attempts` or something so the user could limit the number of attempts if they wish.

I think it might be useful to randomize the order of the regions, especially if you want to run multiple test jobs concurrently.

rjschwei: Fair, I don't think we would need another parameter. The user can limit the number of attempts by setting the regions to try. I am OK with randomizing. We somehow need knowledge of the previously tried and failed region to avoid trying that a second time.

test/unit/services/testing/gce_job_test.py
```diff
@@ -21,6 +21,7 @@ def setup(self):
                 }
             },
             'tests': ['test_stuff'],
+            'test_fallback_regions': [ 'us-west1' ],
```
smarlowucf: No space after '[' or before ']' for formatting.

jgleissner: Seems I didn't run flake8 on the test directory, sorry. Fixed.

smarlowucf:
Fixed test to use `assert_called_once_with` the `assert_called_once` function fails on 3.5.

smarlowucf:
Rebase to fix merge conflicts.

PR #490, opened 2019-07-19
Add guest os features option for gce.

### LINKED ISSUE #487: support GCE "guestOSFeatures"
We need to be able to set guest OS feature in the jobdoc to allow us to indicate UEFI support in our images and other attributes such as blkmq

smarlowucf:

A list of strings can be supplied for a gce image upload.

Unit and integration tests.

### Additional Information

Fixes #487


rjschwei:
Thanks for getting this done in a hurry

PR #492, opened 2019-07-22
Split up API schema

smarlowucf:

Organize and push split schema changes to master. This had cursory review in https://github.com/SUSE-Enceladus/mash/tree/api-auth.

- Simplify schema non empty string. No need for reference instead use dictionary value.
- Split up and reorganize the api schema.
- Add examples to all schema fields.

Unit + integration tests.


smarlowucf:
Re-based to fix ec2 billing code example.

PR #500, opened 2019-07-30
Disable flask-restplus swagger doc.

smarlowucf:

(Merge changes to api-auth feature branch)

Generate static swagger api doc endpoint instead of using flask-restplus.

Unit and integration tests.


mash/services/api/__init__.py
```diff
@@ -32,10 +34,12 @@
 app = Flask(__name__, static_url_path='/static')
 api = Api(
     app,
-    version='1.0',
+    version='3.3.0',
     title='MASH API',
-    description='MASH API',
-    validate=True
+    description='MASH provides a set of endpoints for Image Release '
+                'automation into the Public Cloud Frameworks.',
```
rjschwei: s/the//

mash/services/api/__init__.py
```diff
@@ -32,10 +34,12 @@
 app = Flask(__name__, static_url_path='/static')
 api = Api(
     app,
-    version='1.0',
+    version='3.3.0',
```
rjschwei: Why the large version jump?

smarlowucf: This is to sync with the current MASH version.

smarlowucf:
> I take it the swagger files are generated?

No files are generated as this is disabled with `doc=False`.

smarlowucf:
> The swagger files are explicitly licensed with the MIT license. This is incompatible with GPL which covers mash. I'll follow up with legal to see if we can mix code with these licenses.

Ack. I see the license in the artifacts which are pulled from https://github.com/swagger-api/swagger-ui which is Apache licensed. The choice of GPL has caused problems before and I think it's worth a separate discussion. Also, we aren't following the general practices of the Python community with this license choice which goes against our own policy:

> When the project is part of a larger Open Source ecosystem, use the license which is usually used in this ecosystem.

Almost 70% of python packages use a permissive license (MIT, Apache, BSD, etc). https://snyk.io/blog/over-10-of-python-packages-on-pypi-are-distributed-without-any-license/

smarlowucf:
Removed docs endpoint.

smarlowucf:
@rjschwei once this is merged we can merge api-auth branch (poorly named no auth yet) to master. This all coincides with https://github.com/SUSE-Enceladus/mash-client/pull/21.

PR #505, opened 2019-08-06
Check ec2 image status in waiter

### LINKED ISSUE #503: retry on replication failure in EC2
When replication fails to a given region we should try again.

```
ERROR 2019-08-02 19:51:34,751 ReplicationService
     Pass[1]: There was an error replicating image to eu-north-1. Waiter ImageAvailable failed: Max attempts exceeded

ERROR 2019-08-02 20:01:31,311 ReplicationService
     Pass[1]: There was an error replicating image to ap-northeast-3. Waiter ImageAvailable failed: Max attempts exceeded

INFO 2019-08-02 20:01:31,720 ReplicationService
     Deleting job.

ERROR 2019-08-02 20:01:31,720 ReplicationService
     Pass[1]: Error occurred in replication.

WARNING 2019-08-02 20:01:31,722 PublisherService
     Failed upstream.

INFO 2019-08-02 20:01:31,723 PublisherService
     Deleting job.
```

smarlowucf:

Check image status in waiter. Instead of using waiter from boto3. Continue until the image is
marked as failed or available.

Fixes #503

### DIFF
--- a/mash/services/replication/ec2_job.py
+++ b/mash/services/replication/ec2_job.py
@@ -91,7 +91,10 @@ class EC2ReplicationJob(MashJob):
             if reg_info['image_id']:
                 try:
                     self._wait_on_image(
-                        credential, reg_info['image_id'], target_region
+                        credential['access_key_id'],
+                        credential['secret_access_key'],
+                        reg_info['image_id'],
+                        target_region
                     )
                 except Exception as error:
                     self.status = FAILED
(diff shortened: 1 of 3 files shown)


mash/services/replication/ec2_job.py
```diff
@@ -82,13 +83,27 @@ def run_job(self):
                         credential
 
         # Wait for images to replicate, this will take time.
-        time.sleep(600)
+        time.sleep(900)
```
jgleissner: Are we actually sure that an image replication will always take at least 15 mins? I don't think there's a lot of harm in starting to poll earlier, given that it's just one API call per minute. Also, if replication in EC2 fails straight away for some reason, it would be nice if the job didn't hang about for another 15 mins needlessly. Might be an unlikely scenario though.

smarlowucf: Given the change to wait on state a longer sleep here isn't really necessary. However, I think some sort of wait here is okay since even if there is a failure in every region replication it still will take some time.

rjschwei: 5 minutes please, thanks

Yes we know it will take some time before the image copy is complete and therefore I am OK with waiting for a bit before we start poling.

PR #513, opened 2019-08-15
Add handle request utility method.

smarlowucf:

Method will be used for requests to the credentials service api.

### DIFF
--- a/mash/utils/mash_utils.py
+++ b/mash/utils/mash_utils.py
@@ -20,6 +20,7 @@ import datetime
 import json
 import os
 import random
+import requests
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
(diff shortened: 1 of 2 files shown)


mash/utils/mash_utils.py
```diff
@@ -147,3 +149,29 @@ def restart_jobs(job_dir, callback):
...
+    """
+    request_method = getattr(requests, method)
+    data = None if not job_data else JsonFormat.json_message(job_data)
+    uri = ''.join([url, endpoint])
+
+    response = request_method(uri, data=data)
+
+    if response.status_code not in (200, 201):
+        try:
+            msg = response.json()['msg']
```
rjschwei: Do we know that on failure requests always sets the message, I think on failure we should get the information from "response.reason"

smarlowucf: I think this falls under the same situation which was discussed about not validating internal messages. The requests are all internal and responses have a default json structure.

The try catch around this catches any unexpected formats and yields a default message.

smarlowucf: If we want to include code or reason in the exception case message that seems good. But the expected case will have a more verbose message in the `msg` key than "Not Found" or "OK" etc.

PR #532, opened 2019-09-05
Migrate API to app factory model.

smarlowucf:

This prevents issues with circular imports and allows for modular plugin of config objects.


rjschwei:
joe is on vacation, merging

PR #536, opened 2019-09-06
Move user model utils to a separate module.

smarlowucf:


smarlowucf:
Rebased to fix conflicts.

PR #561, opened 2019-09-25
Migrate credentials service config to api.

smarlowucf:

Credentials service api is served by apache and no longer a systemd unit.


smarlowucf:
> spec file should now have a dependency on apache

This is already the case:

```
Requires:       apache2
Requires:       apache2-mod_wsgi-python3
```

PR #564, opened 2019-09-27
API Authentication and Credentials Service

smarlowucf:

- Add authentication to API
- Move Account handling from datastore to database
- Simplify credentials service, only stores/retrieves credentials as encrypted json files
- Convert credentials service to an API vs an AMQP based service
- Add full set of API endpoints for account, user, token and auth handling
- Setup simple API endpoints for job handling (to be built upon)

Unit tests, integration tests and e2e testing with a job in each cloud.


smarlowucf:
All code has been reviewed, this PR is for tracking purposes.

PR #570, opened 2019-10-01
New service raw image uploader

jgleissner:
This adds a service to upload raw image files to a specified location. Only supported job type at the moment supports uploading to an S3 bucket using the credentials from a specified AWS account.

### DIFF
--- a/mash/services/raw_image_uploader_service.py
+++ b/mash/services/raw_image_uploader_service.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2019 SUSE Software Solutions Germany GmbH. All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# mash is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with mash.  If not, see <http://www.gnu.org/licenses/>
+#
+import logging
+import sys
+import traceback
+
+# project
+from mash.mash_exceptions import MashException
+from mash.services.listener_service import ListenerService
+
+
+def main():
+    """
+    mash - uploader service application entry point
+    """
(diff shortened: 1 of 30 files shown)


test/unit/services/raw_image_uploader/config_test.py
```diff
@@ -0,0 +1,21 @@
...
+from test.unit.test_helper import patch_open
+
+from mash.mash_exceptions import MashConfigException
+from mash.services.raw_image_uploader.config import RawImageUploaderConfig
+
+
+class TestRawImagerUploaderConfig(object):
+    def setup(self):
+        self.config = RawImageUploaderConfig(
+            '../data/mash_config.yaml'
```
smarlowucf: These should be updated to absolute path. `test/data/mash_config.yaml`

mash/services/raw_image_uploader_service.py
```diff
@@ -0,0 +1,57 @@
...
+    try:
+        logging.basicConfig()
+        log = logging.getLogger('MashService')
+        log.setLevel(logging.DEBUG)
+        # run service, enter main loop
+        ListenerService(
+            service_exchange='raw_image_uploader',
+            custom_args={
+                'listener_msg_args': ['cloud_image_name', 'image_file'],
+                'status_msg_args': ['cloud_image_name', 'source_regions']
```
smarlowucf: 'cloud_image_name' is already a default in `status_msg_args` https://github.com/SUSE-Enceladus/mash/blob/master/mash/services/listener_service.py#L70

mash/services/testing_service.py
```diff
@@ -38,8 +38,8 @@ def main():
         ListenerService(
             service_exchange='testing',
             custom_args={
-                'listener_msg_args': ['cloud_image_name', 'source_regions'],
-                'status_msg_args': ['source_regions']
+                'listener_msg_args': ['cloud_image_name', 'image_file', 'source_regions'],
+                'status_msg_args': ['source_regions', 'image_file', 'cloud_image_name']
```
smarlowucf: Ditto on 'cloud_image_name'

test/unit/services/raw_image_uploader/config_test.py
jgleissner: Sorry, I should have seen that change coming as I actually reviewed it. ;) PR rebased and updated.

mash/services/raw_image_uploader_service.py
jgleissner: Fixed.

PR #571, opened 2019-10-03
Add email whitelist config option.

smarlowucf:

If a list of emails is provided only those emails can be used to create a mash user. Otherwise all emails can be used.


rjschwei:
It would be great to support domain whitelisting i.e. accept "*@.suse.com"

smarlowucf:
> It would be great to support domain whitelisting i.e. accept "*@.suse.com"

Sounds useful. https://github.com/SUSE-Enceladus/mash/issues/574

PR #573, opened 2019-10-07
Inject configuration into services.

smarlowucf:

- Inject configuration into services.
- Remove the need for a config factory.
- And remove all empty service config modules. No need for service specific config unless the service has specific settings.


smarlowucf:
> I like that change. :)

It's always nice to do some code cleanup.

PR #575, opened 2019-10-08
Move send email methods to new class.

smarlowucf:

Methods specific to sending emails move to a new class to make notifications types injectable in the future.

The mash service still builds the notification message content.


smarlowucf:
Rebased to fix conflicts.

PR #580, opened 2019-10-14
Add source_regions to raw image uploader args.

smarlowucf:

Raw image uploader expects and parses a source_regions argument from testing service.


smarlowucf:
> Sorry I missed that one.

Well we both missed this one.

PR #586, opened 2019-10-24
Implement Azure SAS URL upload.

### LINKED ISSUE #515: job attribute validation should depend (more) on last_service
Some of the jobdoc attributes still need to be included, even if they won't be used, in order to satisfy validation.

For example, given an Azure job with `"last_service": "uploader"`, the following validation errors were reported:
```
MashClientException: 'image_description' is a required property
'emails' is a required property
'label' is a required property
'offer_id' is a required property
'publisher_id' is a required property
'sku' is a required property
```

Since `image_description`, `label`, `offer_id`, `publisher_id`, and `sku` are only used in the Azure _publisher_ service, they should not be required in this case.

In addition, `emails` is only used if `"publish_offer": true` _and_ last service is at least `publisher`.

smarlowucf:

- Validate Azure job args based on last service.
- Publishing args are not required for testing or upload only jobs.
- Implement Azure SAS URL upload.
- If a SAS token, container and storage account are provided use Azure sas upload workflow. The image is uploaded with provided sas information to a page blob and job is finished.

Fixes #515 

Unit and E2E testing.

### DIFF
--- a/mash/services/api/utils/jobs/azure.py
+++ b/mash/services/api/utils/jobs/azure.py
@@ -37,7 +37,6 @@ def update_azure_job_accounts(job_doc):
         'destination_resource_group',
         'destination_storage_account'
     )
-
     publisher_args = (
         'emails',
         'label',
@@ -45,19 +44,36 @@ def update_azure_job_accounts(job_doc):
         'publisher_id',
         'sku'
     )
+    sas_args = (
+        job_doc.get('sas_token'),
+        job_doc.get('sas_container'),
+        job_doc.get('sas_storage_account')
+    )
 
-    for attr in attrs:
-        if attr not in job_doc:
-            job_doc[attr] = getattr(cloud_account, attr)
+    if any(sas_args) and not all(sas_args):
+        raise MashJobException(
+            'sas_token, sas_storage_account and sas_container are all '
+            'required for an Azure SAS upload job.'
+        )
 
-    if job_doc['last_service'] in ('publisher', 'deprecation'):
-        for arg in publisher_args:
-            if arg not in job_doc:
-                raise MashJobException(
-                    'Azure publishing jobs require a(n) '
-                    ' {arg} argument in the job document.'.format(
-                        arg=arg
+    if job_doc.get('sas_token'):
+        if job_doc['last_service'] != 'uploader':
+            raise MashJobException(
+                'Azure uploads using SAS token expect a last service of uploader.'
+            )
+    else:
+        for attr in attrs:
+            if attr not in job_doc:
+                job_doc[attr] = getattr(cloud_account, attr)
+
+        if job_doc['last_service'] in ('publisher', 'deprecation'):
+            for arg in publisher_args:
+                if arg not in job_doc:
+                    raise MashJobException(
+                        'Azure publishing jobs require a(n) '
+                        ' {arg} argument in the job document.'.format(
+                            arg=arg
+                        )
                     )
-                )
 
     return job_doc
--- a/mash/services/api/utils/jobs/azure.py
+++ b/mash/services/api/utils/jobs/azure.py
@@ -16,6 +16,7 @@
 # along with mash.  If not, see <http://www.gnu.org/licenses/>
 #
 
+from mash.mash_exceptions import MashJobException
 from mash.services.api.utils.users import get_user_by_username
 from mash.services.api.utils.accounts.azure import get_azure_account_by_id
 
@@ -27,7 +28,7 @@ def update_azure_job_accounts(job_doc):
     user = get_user_by_username(job_doc['requesting_user'])
     cloud_account = get_azure_account_by_id(job_doc['cloud_account'], user.id)
 
-    attrs = [
+    attrs = (
         'region',
         'source_container',
         'source_resource_group',
@@ -35,10 +36,28 @@ def update_azure_job_accounts(job_doc):
         'destination_container',
         'destination_resource_group',
         'destination_storage_account'
-    ]
+    )
+
+    publisher_args = (
+        'emails',
+        'label',
+        'offer_id',
+        'publisher_id',
+        'sku'
+    )
 
     for attr in attrs:
         if attr not in job_doc:
             job_doc[attr] = getattr(cloud_account, attr)
+
+    if job_doc['last_service'] in ('publisher', 'deprecation'):
+        for arg in publisher_args:
+            if arg not in job_doc:
+                raise MashJobException(
+                    'Azure publishing jobs require a(n) '
+                    ' {arg} argument in the job document.'.format(
+                        arg=arg
+                    )
+                )
 
     return job_doc
--- a/mash/services/jobcreator/azure_job.py
+++ b/mash/services/jobcreator/azure_job.py
@@ -53,6 +53,9 @@ class AzureJob(BaseJob):
         self.offer_id = self.kwargs.get('offer_id')
         self.publisher_id = self.kwargs.get('publisher_id')
         self.sku = self.kwargs.get('sku')
+        self.sas_token = self.kwargs.get('sas_token')
+        self.sas_container = self.kwargs.get('sas_container')
+        self.sas_storage_account = self.kwargs.get('sas_storage_account')
 
     def get_deprecation_message(self):
         """
@@ -165,17 +168,29 @@ class AzureJob(BaseJob):
         """
         Build uploader job message.
         """
-        uploader_message = {
-            'uploader_job': {
-                'cloud_image_name': self.cloud_image_name,
-                'cloud': self.cloud,
-                'account': self.cloud_account,
-                'region': self.region,
-                'container': self.source_container,
-                'resource_group': self.source_resource_group,
-                'storage_account': self.source_storage_account
+        if self.sas_token:
+            uploader_message = {
+                'uploader_job': {
+                    'cloud_image_name': self.cloud_image_name,
+                    'cloud': self.cloud,
+                    'container': self.sas_container,
+                    'storage_account': self.sas_storage_account,
+                    'sas_token': self.sas_token
+                }
             }
-        }
+        else:
+            uploader_message = {
+                'uploader_job': {
+                    'cloud_image_name': self.cloud_image_name,
+                    'cloud': self.cloud,
+                    'account': self.cloud_account,
+                    'region': self.region,
+                    'container': self.source_container,
+                    'resource_group': self.source_resource_group,
+                    'storage_account': self.source_storage_account
+                }
+            }
+
         uploader_message['uploader_job'].update(self.base_message)
 
         if self.cloud_architecture:
--- a/mash/services/jobcreator/azure_job.py
+++ b/mash/services/jobcreator/azure_job.py
@@ -32,11 +32,6 @@ class AzureJob(BaseJob):
         """
         try:
             self.cloud_account = self.kwargs['cloud_account']
-            self.emails = self.kwargs['emails']
-            self.label = self.kwargs['label']
-            self.offer_id = self.kwargs['offer_id']
-            self.publisher_id = self.kwargs['publisher_id']
-            self.sku = self.kwargs['sku']
         except KeyError as error:
             raise MashJobCreatorException(
                 'Azure jobs require a(n) {0} key in the job doc.'.format(
@@ -53,6 +48,11 @@ class AzureJob(BaseJob):
         self.destination_container = self.kwargs.get('destination_container')
         self.destination_resource_group = self.kwargs.get('destination_resource_group')
         self.destination_storage_account = self.kwargs.get('destination_storage_account')
+        self.emails = self.kwargs.get('emails')
+        self.label = self.kwargs.get('label')
+        self.offer_id = self.kwargs.get('offer_id')
+        self.publisher_id = self.kwargs.get('publisher_id')
+        self.sku = self.kwargs.get('sku')
 
     def get_deprecation_message(self):
         """
--- a/mash/services/api/schema/jobs/azure.py
+++ b/mash/services/api/schema/jobs/azure.py
@@ -57,6 +57,13 @@ azure_job_message['properties']['destination_storage_account'] = string_with_exa
     'storage_account2'
 )
 azure_job_message['properties']['region'] = string_with_example('westus')
+azure_job_message['properties']['sas_token'] = string_with_example(
+    'sv=2015-07-08&sr=b&sig=123456789&se=2016-10-18T21&sp=row'
+)
+azure_job_message['properties']['sas_container'] = string_with_example('container1')
+azure_job_message['properties']['sas_storage_account'] = string_with_example(
+    'storage_account1'
+)
 
 azure_job_message['required'].append('cloud_account')
 azure_job_message['properties']['image']['example'] = \
--- a/mash/services/api/schema/jobs/azure.py
+++ b/mash/services/api/schema/jobs/azure.py
@@ -58,11 +58,6 @@ azure_job_message['properties']['destination_storage_account'] = string_with_exa
 )
 azure_job_message['properties']['region'] = string_with_example('westus')
 
-azure_job_message['required'].append('emails')
-azure_job_message['required'].append('label')
-azure_job_message['required'].append('offer_id')
-azure_job_message['required'].append('publisher_id')
-azure_job_message['required'].append('sku')
 azure_job_message['required'].append('cloud_account')
 azure_job_message['properties']['image']['example'] = \
     'openSUSE-Leap-15.0-Azure'
--- a/mash/services/api/wsgi.py
+++ b/mash/services/api/wsgi.py
@@ -4,4 +4,4 @@ from mash.services.api.config import Config
 application = create_app(Config())
 
 if __name__ == '__main__':
-    application.run(port=5000)
+    application.run(port=5005)
--- a/mash/services/uploader/azure_job.py
+++ b/mash/services/uploader/azure_job.py
@@ -39,10 +39,7 @@ class AzureUploaderJob(MashJob):
         self.cloud_image_name = ''
 
         try:
-            self.account = self.job_config['account']
-            self.region = self.job_config['region']
             self.container = self.job_config['container']
-            self.resource_group = self.job_config['resource_group']
             self.storage_account = self.job_config['storage_account']
             self.base_cloud_image_name = self.job_config['cloud_image_name']
         except KeyError as error:
@@ -53,6 +50,11 @@ class AzureUploaderJob(MashJob):
                 )
             )
 
+        self.account = self.job_config.get('account')
+        self.region = self.job_config.get('region')
+        self.resource_group = self.job_config.get('resource_group')
+        self.sas_token = self.job_config.get('sas_token')
+
     def run_job(self):
         self.status = SUCCESS
         self.send_log('Uploading image.')
@@ -61,21 +63,63 @@ class AzureUploaderJob(MashJob):
             self.base_cloud_image_name
         )
 
-        self.request_credentials([self.account])
-        credentials = self.credentials[self.account]
         blob_name = ''.join([self.cloud_image_name, '.vhd'])
 
+        if self.sas_token:
+            # Jobs with sas_tokens are upload only
+            self._upload_image(blob_name)
+            self.send_log(
+                'Uploaded blob: {blob} using sas token.'.format(
+                    blob=blob_name
+                )
+            )
+        else:
+            self.request_credentials([self.account])
+            credentials = self.credentials[self.account]
+
+            self._upload_image(blob_name, credentials)
+            self._create_image(blob_name, credentials)
+
+            self.source_regions[self.region] = self.cloud_image_name
+            self.send_log(
+                'Uploaded image has ID: {0} in region {1}'.format(
+                    self.cloud_image_name,
+                    self.region
+                )
+            )
+
+    @property
+    def image_file(self):
+        """System image file property."""
+        return self._image_file
+
+    @image_file.setter
+    def image_file(self, system_image_file):
+        """
+        Setter for image_file list.
+        """
+        self._image_file = system_image_file
+
+    def _upload_image(self, blob_name, credentials=None):
+        """
+        Upload image as a page blob to an ARM container.
+
+        Run upload in a separate process as zero page check
+        is CPU intensive. Raise exception if result queue
+        is not empty.
+        """
         result = SimpleQueue()
         args = (
             blob_name,
             self.container,
-            credentials,
             self.image_file,
             self.config.get_azure_max_retry_attempts(),
             self.config.get_azure_max_workers(),
-            self.resource_group,
             self.storage_account,
-            result
+            result,
+            credentials,
+            self.resource_group,
+            self.sas_token
         )
         upload_process = Process(
             target=upload_azure_image,
@@ -87,6 +131,10 @@ class AzureUploaderJob(MashJob):
         if result.empty() is False:
             raise MashUploadException(result.get())
 
+    def _create_image(self, blob_name, credentials):
+        """
+        Create image in ARM from existing page blob.
+        """
         with create_json_file(credentials) as auth_file:
             compute_client = get_client_from_auth_file(
                 ComputeManagementClient, auth_path=auth_file
@@ -112,23 +160,3 @@ class AzureUploaderJob(MashJob):
                 }
             )
             async_create_image.wait()
-
-        self.source_regions[self.region] = self.cloud_image_name
-        self.send_log(
-            'Uploaded image has ID: {0} in region {1}'.format(
-                self.cloud_image_name,
-                self.region
-            )
-        )
-
-    @property
-    def image_file(self):
-        """System image file property."""
-        return self._image_file
-
-    @image_file.setter
-    def image_file(self, system_image_file):
-        """
-        Setter for image_file list.
-        """
-        self._image_file = system_image_file
--- a/mash/utils/azure.py
+++ b/mash/utils/azure.py
@@ -71,8 +71,8 @@ def copy_blob_to_classic_storage(
     """
     Copy a blob from ARM based storage account to a classic storage account.
     """
-    source_page_blob_service = get_page_blob_service(
-        auth_file, source_resource_group, source_storage_account
+    source_page_blob_service = get_page_blob_service_with_account_keys(
+        auth_file, source_storage_account, source_resource_group
     )
 
     source_blob_url = get_blob_url(
@@ -110,8 +110,8 @@ def delete_page_blob(
     """
     Delete page blob in container.
     """
-    page_blob_service = get_page_blob_service(
-        auth_file, resource_group, storage_account
+    page_blob_service = get_page_blob_service_with_account_keys(
+        auth_file, storage_account, resource_group
     )
     page_blob_service.delete_blob(container, blob)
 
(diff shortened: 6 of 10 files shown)


mash/services/api/utils/jobs/azure.py
```diff
@@ -27,18 +28,52 @@ def update_azure_job_accounts(job_doc):
...
         'region',
         'source_container',
         'source_resource_group',
         'source_storage_account',
         'destination_container',
         'destination_resource_group',
         'destination_storage_account'
-    ]
+    )
+    publisher_args = (
```
rjschwei: Shouldn't this be `publishing_args` as in the args used for publishing?

mash/services/api/utils/jobs/azure.py
```diff
@@ -27,18 +28,52 @@ def update_azure_job_accounts(job_doc):
...
+        raise MashJobException(
+            'sas_token, sas_storage_account and sas_container are all '
+            'required for an Azure SAS upload job.'
+        )
+
+    if job_doc.get('sas_token'):
+        if job_doc['last_service'] != 'uploader':
+            raise MashJobException(
+                'Azure uploads using SAS token expect a last service of uploader.'
+            )
```
rjschwei: This breaks the idea that we can upload an image that gets published as an image and have the same file uploaded to a container. That's a workflow we support for the other use case. We should support it in this case as well.

mash/services/jobcreator/azure_job.py
```diff
@@ -165,17 +168,29 @@ def get_uploader_message(self):
...
+                    'cloud_image_name': self.cloud_image_name,
+                    'cloud': self.cloud,
+                    'account': self.cloud_account,
+                    'region': self.region,
+                    'container': self.source_container,
+                    'resource_group': self.source_resource_group,
+                    'storage_account': self.source_storage_account
+                }
+            }
+
```
rjschwei: This should not be either or. We should be able to use the same file for upload to a SAS URL and turn it into an image.

mash/services/api/utils/jobs/azure.py
smarlowucf: The service is publisher the action is publishing. In code we always use publisher so I think it's okay to stick with the same convention.

mash/services/jobcreator/azure_job.py
smarlowucf: IMHO we should solve the problem at hand and not worry about what other problem there might be in the future. We can guess at what needs there might be in the future but they might not be there when the future arrives. If there comes a need to upload to a SAS URL that becomes an image in the framework then that is better tackled when the need arrives.

bear454: We cannot extend the workflow from upload to a SAS URL; it's a final destination. In order to register an image from a blob in storage, a lease is created on the backing blob; this cannot be done with a SAS URL.

smarlowucf:
Any follow up here? Otherwise will move forward.

rjschwei:
The raw image uploader service was implemented to cover this case and the plan was to expand it when we need the same functionality for other frameworks as we started out with EC2. We should stick to the architectural decision made at the time rather than pushing code into other places for convenience.

smarlowucf:
> The raw image uploader service was implemented to cover this case and the plan was to expand it when we need the same functionality for other frameworks as we started out with EC2.

The workflow for raw image uploader doesn't fit this use case. The requirement I was given was that we need an image uploaded to a SAS URL and we don't want to run testing because it's specifically for the Azure team to do their own testing. Given this it doesn't make sense to force the job through uploader and testing to get to raw image uploader. Also, this code while it's slightly different to upload to an SAS URL is very similar to a page blob upload to a container. It fits very nicely in the uploader.



> We should stick to the architectural decision made at the time rather than pushing code into other places for convenience.

If the workflow doesn't fit we shouldn't force it. And we should always be willing to pivot if a previous decision doesn't make sense anymore or in all cases.

rjschwei:
Let's see if we can agree that this is a raw image upload?

smarlowucf:
> Let's see if we can agree that this is a raw image upload?

It could be considered a raw image upload. However, the order of the upload is key here. I have one user who requested a workflow where a raw image upload can be added as part of a publishing job. The other user requested a raw image upload that could be done as a "one-off" job. I think both requests are perfectly valid use cases of MASH. I understand that this means the functionality between Azure and EC2 acts differently (for now) but the end goal is that you can do either workflow in any cloud and to any supported provider. This end goal is something that we haven't achieved yet.

When the underlying pipeline becomes less opinionated the users will have more flexibility. Then the implementation in this pull request can be directly used in Azure for either workflow which I think solves your concerns.

jgleissner:
FWIW, if the SAS upload is a dead-end (as in you can't progress to testing and publishing) it feels misplaced in the uploader service to me. I realize there is a use case for just uploading an image and not doing the image creation and testing steps before that, but I guess there's also the use case that you do want to test the image before uploading to an SAS URL, and if I understand correctly, that's not possible with this implementation.

Implementing SAS upload as a raw image upload would allow to go through the test cycle, so it seems more flexible to me. It should be possible to support skipping the upload/test steps, if you only want to do the SAS upload and nothing else.

In fact, from a dev and testing point of view, it would be convenient if you could skip steps. For example, for testing raw image uploader, I placed fake job messages in `/var/lib/mash/*jobs/` and sent a crafted message into the queue to not have to go through the download/upload/testing steps every time, which wasn't exactly convenient (perhaps there's an easier way?).

smarlowucf:
I think we are barking up the same tree here. In the end the most user friendly option is flexibility right?

- Maybe I want to upload/test an image in Azure then upload it to a bucket with a SAS token
- Or maybe I just want to upload a tarball to S3
- Or maybe I want to do an S3 upload as part of a publishing job

All of these to me are valid use cases and handling them all seems like a good idea.

I agree that an uploader service workflow that doesn't create an image right now doesn't make complete sense. This is because we've kind of married two different concepts into one service. However, this can be split out and we could have an uploader service and an image creation service. That way uploader service does one thing great and that's upload a tarball to a set of providers. The image creation service is only tasked with creating images in certain frameworks.

The idea of being able to skip steps falls right into place with a workflow where one service is tasked with managing the flow of jobs. I think this is a good idea but we aren't there yet. Right now we have a very opinionated set of chained services that must be called in a specific order.

smarlowucf:
I think we are having a great conversation about where MASH should get to ideally. And all use cases discussed are viable and should eventually be accounted for. However, we should focus on one step at a time. We cannot solve the world in one pull request. I think that every implementation for SAS upload right now is going to have it's issues or concerns. I think there's a clear path with how we can solve all these concerns but it's going to take time. Are we amenable to moving the architecture discussion to a better platform?

domvieira:
To weigh in, I like having this high-level discussion here (although it also should happen elsewhere). It creates a lot of shared context. I think @jgleissner makes some excellent points above and @smarlowucf , it sounds like you're in general agreement in terms of the user flexibility that implementing SAS upload as a "raw image upload" can provide. Outside of separating all the services and the heavy lifting required to implement all the use cases as noted, can we do a simple effort here to see how running it as a raw image upload would work? That would help us move a few steps towards the more flexible design.

rjschwei:
I agree that we cannot boil the ocean. However there is really no need to marry the SAS raw image target upload with the existing uploader/image creator. The shared code between the current uploader and that is required for the SAS target upload can be factored out and then used in the uploader/creator and the raw image uploader.

Put the raw image uploader after the OBS service and then SAS upload jobs get `last_service: raw_image_uploader`. Then we satisfy all the requirements we know of today which are:

- Upload the raw image file to a target where a third party can pick it up and use the same image file for image creation
- Upload the raw image file to a target where a third party can pick

Yes notice the Azure requirement and the requirement for the CHOST image start with the exact same wording. It is a matter of arranging the pipeline that allows us to avoid special cases.

Leaves the question of do we want to clean up a raw image that has been uploaded if that job also included image creation and the test fails? We can decide that as we see fit.

smarlowucf:
To @domvieira comment. I think this PR does a good job of showing how we can solve the SAS upload with the uploader service. I think we've done a good job covering all the advantages and disadvantages.

If we were to use the raw image uploader service and try to do so with minimal changes the following steps are my best guess at what would be needed:

- We would need a way to skip uploading and testing for these jobs. Right now we have a IMHO bad way to do this with specific clouds. For example GCE has a "none" implementation of replication so that step is essentially skipped. However, that doesn't work here because we cannot simply skip a job based on the cloud and the architecture of the pipeline is driven by cloud framework. This is not ideal but it's where we are at based on having an opinionated pipeline message flow. The simplest option is adding another flag to the schema that can be set for each service. "skip_upload", "skip_testing".
- Once we can skip upload we need to modify the schema and validation for all cloud frameworks because upload parameters are now optional since we are keying on raw image uploader args. The entire workflow of MASH hinges on the fact that obs and uploader **must** be part of every job. Removing this built in requirement is not trivial. The idea that you must do at least an upload is something that was agreed upon very early on. And that's why it also wouldn't be easy to simply lift and shift the raw_image_uploader into it's place.
- With these things in place we can implement an Azure based raw image uploader class that handles the underlying upload.

IMHO this route is much more intrusive because it affects every single service class for every cloud framework. It also means modifying every validation schema. This fix is Ideally temporary anyway. Should we really choose the more complex change as a temporary solution if in the end it becomes deprecated?

domvieira:
Thanks for the explanation @smarlowucf . Appreciate the perspective. I'll let others weigh in but can you explain a bit more on your ending comment? Why is adding this feature "ideally temporary"? Is there a case where this feature wouldn't be needed?

> This fix is Ideally temporary anyway. Should we really choose the more complex change as a temporary solution if in the end it becomes deprecated?

smarlowucf:
@rjschwei When you mention moving the raw image uploader it reminds me why we have some of this confusion. If We went the route you explain it's the exact same thing Joe and I are talking about except for naming semantics.

If you rename raw image uploader to uploader and uploader to image creator then we are talking about the same solution.

The reason for the confusion is that when I implemented mash workflow I did EC2 first. EC2 doesn't really have a typical upload step. It's a rather complex upload process that requires a running instance and leads directly to an image or snapshot. This lead to Azure and GCE uploader implementing upload and image creation.

So no matter if we call it raw image uploader or uploader for short all it would do is upload tarballs somewhere. The "uploader" in your workflow really becomes an image creation service. For Azure and GCE there's two concrete steps to upload and create an image. For EC2 it's a bit muddled and therefore it would have a "None" implementation for "raw image uploader".

Maybe that helps us come closer to an agreement?

rjschwei:
What it is called in somewhat semantic, but more on that below. Let's look at what we have in azure_job.py in the uploader service today. There one will find  `async_create_image = compute_client.images.create_or_update(`. Meaning the "uploader" (as it is named today) implementation for Azure follows the semantics of AWS where cloud image creation, i.e. the thing a user can launch an instance from, is done in the same logical code unit as the uploading of the image file. This logical code unit could also be called "image_creator"

As I stated previously we have two use cases with almost identical requirements. Upload an image file to $SOME_LOCATION is the requirement shared for the CHOST image and for the Lisa Azure test case. The CHOST image use case goes father in that it adds "use the same file to create a cloud image". For the CHOST image we decided that $SOME_LOCATION is an S3 bucket. For Lisa we want $SOME_LOCATION to be an Azure container hidden behind a SAS URL.

For uploading a raw image file, i.e. put an image file into $SOME_LOCATION, we decided to implement a new service in the pipeline called "raw_image_uploader" #570  and the initial upload target was implemented as an S3 bucket to cover the CHOST image use case. This, the "raw_image_uploader", is the logical code unit that puts a raw image file into $SOME_LOCATION. Since the Lisa use case is exactly that, put the raw image file into a specific target, from a code logical point of view SAS URL upload should live in "raw_image_uploader".

That in #570 it was decided to place the "raw_image_uploader" service after image creation is at issue here. Because of this the notion of "we have to skip something" is floating around.

The raw image uploading functionality for Azure already lives in a utility (mash.utils.azure). We can re-use this in the "raw_image_uploader" and put the vhdfixed file into the Azure container that feeds Lisa.

We also need to move "raw_image_uploader" from fourth in the pipeline to second. Once this is done we have the same functionality for all frameworks. A given image file can be uploaded to either an Azure container hidden behind a SAS URL or an S3 bucket and that same file can progress down the mash pipeline and become a cloud image.

This is very useful in that for the first time we could have the exact same image, on the day of release, run through the full Lisa test suite (albeit the results don't gate the release) and become a cloud image in Azure MP. The release is still gated by img-proof as it is today. Further this approach gives us full flexibility.

With the "raw_image_uploader" #570 placed second in the pipeline it becomes a no op if the job doc does not include the information for $SOME_TARGET. Any given image can be uploaded as a raw image to $SOME_TARGET and the same image can turn into a cloud image as it gets processed farther down the pipeline. For the Azure case that affords us the flexibility of 4 cases:

- Run Lisa and have an image release from the exact same image
- Run Lisa only `last_service: raw_image_uploader`
- Run Lisa and img-proof `last_service: testing`
- Skip Lisa, do not supply the SAS url info in the job doc

In order to achieve this the code we call the "uploader" today stays exactly as it is. If it is preferred to re-name this logical code unit to "image_creator" for clarity then let's do that. The "raw_image_uploader" #570 moves to the second place in the pipeline. The code that needs touching to make this work is primarily the "raw_image_uploader" code.

If we do not keep the same semantics that the code that creates the image currently called "uploader" does it's own uploading then we loose the flexibility to have $SOME_TARGET being disconnected from the account/subscription that is used to create the cloud image, or we have to invent some looping mechanism in something that was designed to be linear. This will create other problems. So we want to keep the semantics of the "uploader" as they are today and again if a rename to "image_creator" is desired fine by me.

This is a solution that affords us the most flexibility, is relatively un-intrusive as the bigger code changes are focused on the logical code unit of "raw_image_uploader". Additional code changes will be necessary in the job doc schema and processing. Also this gives us the same logical functionality for all frameworks, 1 image file can be uploaded in raw format and can become a cloud image. We have a choice if the raw image file ends up in an S3 bucket or in an Azure container.

smarlowucf:
@domvieira Yes, part of this discussion is happening in an email thread so some context is not coming across. Other context I have not fully fleshed out because it is an architectural vision which I have not put on paper yet.

First lets ignore everything we know about MASH right now. If I upload an image to S3 or to a bucket in Azure or to Dropbox or <enter any service> it's basically a raw image upload right? We are taking a tarball and uploading it somewhere. Is this agreeable terminology?

I think the confusion is coming because I named the second service in MASH "uploader". This is misleading and it was a poor design decision with only EC2 in mind. Right now the uploader service for GCE and Azure actually has two independent steps. It does a "raw image" upload to a bucket. Then it does a framework image creation. In EC2 there's not really a way to upload tarballs directly so "upload" is really two steps in one call to ec2imgutils. This one call performs a large number of steps which includes uploading the image to a volume and eventually creating a snapshot that becomes a framework image.

This is where my premise that "raw image uploader" is duplicative comes from. The problem stems from my decision to include a "raw image" upload step in the uploader service for GCE and Azure. A good design would eliminate duplicative code and would separate independent tasks into separate services.

Thus when I say "uploader" I really mean "raw image" uploader. And this uploader service would be paired down to only do tarball uploads. Image creation would then become a separate step. For EC2 since image creation is tied to the upload process it's implementation is stuck together in the new image creation service. So "uploader" and "raw image" uploader end up one and the same.

In this new workflow we would have a controller "job creator" that handles the flow of a job through the pipeline. Instead of the flow being forced by the design it would be flexible. Thus we could account for use cases that have been discussed without the need for "None" implementations.

I.e. for the EC2 publishing job where we also want to do a raw image upload then the controller knows the flow should be:

OBS -> uploader -> image creator -> testing -> uploader (with S3 info) -> replication -> publisher -> deprecation

For an Azure one-off upload the flow is:

OBS -> uploader (with SAS token)

For an Azure raw image upload that is tested:

OBS -> uploader -> image creator -> testing -> uploader (with SAS token)

For a GCE publishing job we now get to skip services that aren't required:

OBS -> uploader -> image creator -> testing -> deprecation

All of the business rules or conditionals that determine the flow are in one spot (new controller) and are easily determined by the user which means the user is given flexibility while simplifying a lot of the service code.

This change in flow not only solves our problem here with different image upload use cases but it also solves:

- Cross framework sync (the controller can handle and compare multiple images before sending the jobs to uploader)
- Job status in API (the controller will now be getting constant feedback about a job which it can make available to the API)
- It removes embedded knowledge about the flow of jobs that each service currently holds (E.g. each service right now is "hard coded" with it's previous and next service. It uses this to know where to get and send messages. Needless to say this isn't flexible.)
- It provides a cleaner way for MASH to do retries. (instead of embedding retries in every service it can be in one spot in a controller)
- It greatly simplifies debugging and testing since you could then start a job at any point without stubbing services and manually creating/adding message to the queue.
- It simplifies the initial message flow as there's no need for a listener message (result from prev service) and a service message (job initialization). It would be one simple message.
- It removes the need for "None" classes because the controller would know that for example GCE doesn't have replication or publishing.
- Probably other great things I haven't thought of.

Does this help to clarify the thinking? I can work on a full proposal with message diagrams if need be.

smarlowucf:
> Since the Lisa use case is exactly that, put the raw image file into a specific target, from a code logical point of view SAS URL upload should live in "raw_image_uploader".

I agree. But the GCE and Azure upload steps that are currently in the wrong service are also raw image uploads. And what if I want to do a raw image upload to a bucket in GCE? Isn't this the exact same thing as it is in the process of a publishing job? If we distinguish raw image upload from upload then we are either duplicating the same code twice or we are being opinionated and saying you can only upload to certain places and raw image upload to other places. To me it makes no sense. These are exactly the same thing.

> If we do not keep the same semantics that the code that creates the image currently called "uploader" does it's own uploading then we loose the flexibility to have $SOME_TARGET being disconnected from the account/subscription that is used to create the cloud image, or we have to invent some looping mechanism in something that was designed to be linear. This will create other problems.

Disagree. The workflow I'm suggesting still has an implicit linear set of steps. The flow is controlled by a "controller" service which means it can be more flexible. I.e. we no longer have to force GCE jobs through replication and publishing when really these services should know nothing about GCE. It also means that we can simplify all "raw image" uploading to one service. And the controller knows when and with what parameters to call the uploader. It also means we can start jobs anywhere in the pipeline. I understand the original decision was to not allow this but it's very heavy handed and it makes debugging and testing extremely and unnecessarily difficult. I think @jgleissner  mentioned this exact issue as well in a different thread.

> This is a solution that affords us the most flexibility

Also disagree, the controller workflow I think provides more flexibility. Especially if you consider the idea that the order of operations might change again in the future. Since we move knowledge of the order of operations from every service down to one controller service it will be much easier to make changes in the future. And if we split out all "raw image" uploading to one service then it means we have the flexibility to use any supported service for "raw image" upload step. E.g. if these are separate services they have separate functions:

- raw image uploader
  - S3
  - Azure bucket via SAS token
- uploader
  - Azure bucket via storage account keys
  - GCE bucket
  - ec2imgutils (moves to image creator since there's no real raw image upload step)

- combined raw image uploader
  - S3
  - Azure bucket via SAS token or storage account keys
  - GCE Bucket
  - Anything else

Of course the controller will know that image upload step must be a subset of these but it gives the flexibility without code duplication to do a raw image upload step to any of these options. You don't get this flexibility if you keep the current "uploader" steps and "raw image" uploader steps segregated unless you are willing to duplicate code.

rjschwei:
Before I start answering let clarify/amend what I stated previously. It is clear that a loop with fork could be inserted into the code that does the raw image file upload. However that idea that the same image progresses to multiple places from the same code stream was rejected. See my initial request for changes.

rjschwei:
> > Since the Lisa use case is exactly that, put the raw image file into a specific target, from a code logical point of view SAS URL upload should live in "raw_image_uploader".
> 
> I agree. But the GCE and Azure upload steps that are currently in the wrong service are also raw image uploads. And what if I want to do a raw image upload to a bucket in GCE? Isn't this the exact same thing as it is in the process of a publishing job?

Yes it is, but we do not have a request to satisfy GCE ATM. Here is a difference between "just over the edge of the plate" and "boiling the ocean". Re-arranging the Azure code if it is now amenable to have a for in the code is closely related and triggered by a feature request, while touching the GCE code would be touching code that is logically related but doesn't have to be touched to satisfy Azure.


> If we distinguish raw image upload from upload then we are either duplicating the same code twice 

We would duplicate the setup code, that is true. But setup code i.e. collection of parameters to drive the code that actually does the work would be expected to be small. This could be avoided with forking prior to the setup code, but you already rejected the idea of forking.

> or we are being opinionated and saying you can only upload to certain places and raw image upload to other places. 

The only opinion the code will have is that the, in either implementation BTW is that the raw image upload, i.e. where a 3rd party can pick up the raw image and the location where the raw image gets stored to create a cloud image for instance creation would be different. Albeit this is an "implicit" opinion that we could and would not enforce. If we would parallelize such an upload the user would inevitable end up with data corruption.

The rest of the comments pertain primarily to GCE and larger changes that are not related to "where do we do Azure raw image uploading" and should be considered as part of a different discussion.

smarlowucf:
> Yes it is, but we do not have a request to satisfy GCE ATM. Here is a difference between "just over the edge of the plate" and "boiling the ocean". Re-arranging the Azure code if it is now amenable to have a for in the code is closely related and triggered by a feature request, while touching the GCE code would be touching code that is logically related but doesn't have to be touched to satisfy Azure.

If we updated Azure based on this idea, that the upload code is incorrectly mashed into the uploader service and really it should be two separate steps, it doesn't force us to change GCE right away. The changes submitted in this pull request are one minor step towards a goal.

But if you are arguing that point, then moving the raw image uploader affects every service and every class for all clouds. So that change unnecessarily affects GCE and EC2 whereas the current PR only affects Azure.

> We would duplicate the setup code, that is true. But setup code i.e. collection of parameters to drive the code that actually does the work would be expected to be small. This could be avoided with forking prior to the setup code, but you already rejected the idea of forking.

I don't follow the forking part can you point to this previous request? I don't see forking mentioned earlier in the discussion.

> The rest of the comments pertain primarily to GCE and larger changes that are not related to "where do we do Azure raw image uploading" and should be considered as part of a different discussion.

I thought this turned into an overarching architectural discussion? The decision I made in how to implement this feature was directly related to; solving the problem at hand, making the change based on current requirements, and factoring in future requirements which currently includes one of the next big line items to tackle. I mentioned all of this other information to help you see the vision I have for where we need to go based on specific requirements we have in the todo list.

smarlowucf:
As I think this has devolved into a shouting match of opinions how do we determine one opinion is right over the rest? The route I provided for this PR is not a rash find the simplest fix and move on solution. I've careful considered the current architecture, the details of the feature request and also features that will be coming in the pipeline. The solution is based on ideas/features that I think will be needed in the near future. For example:

- I think we made a bad decision to mash two steps into the current uploader and I think this should slowly be cleaned up.
- I think as part of this cleanup all uploading should be combined into one concrete uploader service.
- I think we have to make architectural changes to the pipeline to handle synchronizing multiple image jobs.
- I think that the current requirement to implement empty classes for services that are not required is poor design and that it should eventually cleaned up.

Obviously I'm biased towards the solution I came to and other are biased to their own ideas. So how can we get this moving forward?

rjschwei:
Well I would classify a shouting match differently. Anyway, we have 2 participants that think the SAS upload belongs into the raw-image-uploader. To get there there is one proposal to implement "skipping" functionality and another proposal to-rearrange the pipeline. I feel pretty confident that we'll find a solution to these different approaches.

rjschwei:
> > Yes it is, but we do not have a request to satisfy GCE ATM. Here is a difference between "just over the edge of the plate" and "boiling the ocean". Re-arranging the Azure code if it is now amenable to have a for in the code is closely related and triggered by a feature request, while touching the GCE code would be touching code that is logically related but doesn't have to be touched to satisfy Azure.
> 
> If we updated Azure based on this idea, that the upload code is incorrectly mashed into the uploader service and really it should be two separate steps, it doesn't force us to change GCE right away. The changes submitted in this pull request are one minor step towards a goal.
> 
> But if you are arguing that point, then moving the raw image uploader affects every service and every class for all clouds. So that change unnecessarily affects GCE and EC2 whereas the current PR only affects Azure.
> 

How so, can you please elaborate on this point?

If we move the raw-image-uploader to position 2, the obs-service would now send it's message to the new service. The target is hard coded in OBS, so yes there is a change. The raw image uploader gets to decide whether to do something or not based on what's in the job-doc. It then sends the exact message it received from obs-service to the uploader. From there nothing else in the pipeline is touched. Therefore:
 - change in obs to send message to raw-img-uploader
 - update the raw-img-uploader code to handle SAS upload
 - Potential update in uploader from where to expect it's message

The changes outside the raw-image-uploader should be 1 location each in code.

> > We would duplicate the setup code, that is true. But setup code i.e. collection of parameters to drive the code that actually does the work would be expected to be small. This could be avoided with forking prior to the setup code, but you already rejected the idea of forking.
> 
> I don't follow the forking part can you point to this previous request? I don't see forking mentioned earlier in the discussion.

Well if we use the same code to upload to two different locations we need a fork. In the current proposal a fork was effectively created by inserting an exception and enforcing that "uploader" has to be `last_service` when SAS upload information is provided.

> 
> > The rest of the comments pertain primarily to GCE and larger changes that are not related to "where do we do Azure raw image uploading" and should be considered as part of a different discussion.
> 
> I thought this turned into an overarching architectural discussion? 

No from my point of view, moving a service in the pipeline to a different slot is not, IMHO and over all architectural discussion.

smarlowucf:
> Anyway, we have 2 participants that think the SAS upload belongs into the raw-image-uploader.

Do you not agree that there is a strong bias by those two participants because they were directly involved in the creation of raw image uploader? Therefore this bias may be swaying their opinion for the proposed route in this PR because they want to see their previous work used in the solution?

Right now the only working proposed solution is the one in this PR. I have not seen a good reason why this is a bad PR. The argument seems to me to come down to "we made a decision before we should not pivot from that decision". But as part the discussion there is a competing argument that's being used in the opposite direction. Sticking with raw image uploader means we will have to change previously agreed upon decisions that were made by other participants. Why is it that decisions made by you can not be changed and decisions made by others are free to be changed, removed or ignored without discussion?

rjschwei:
> > Anyway, we have 2 participants that think the SAS upload belongs into the raw-image-uploader.
> 
> Do you not agree that there is a strong bias by those two participants because they were directly involved in the creation of raw image uploader? 

No, and further, if I recall correctly you were part of the decision to create the raw-image-uploader service and did not voice any concerns. At the time the decision was made it was already pointed out that one could easily envision that we'll have to support other targets for such raw image uploads, including Azure targets.

> Therefore this bias may be swaying their opinion for the proposed route in this PR because they want to see their previous work used in the solution?

I think you are starting to go down an unfavorable path and the implication of ill will by others is not helpful.

> 
> Right now the only working proposed solution is the one in this PR. I have not seen a good reason why this is a bad PR. 

As stated elsewhere no one has stated that this is a "bad PR" or "abhorrent" as you stated in the e-mail conversation. Again this is not really helpful.

> The argument seems to me to come down to "we made a decision before we should not pivot from that decision". But as part the discussion there is a competing argument that's being used in the opposite direction. Sticking with raw image uploader means we will have to change previously agreed upon decisions that were made by other participants. Why is it that decisions made by you can not be changed and decisions made by others are free to be changed, removed or ignored without discussion?

As stated in e-mail one can only consider what one knows. Given private conversation between parties that are not documented in the code means others cannot really know about those decisions. Again the implication of ill will here is not helpful.

smarlowucf:
@rjschwei I was sure to mention previously that:

> Obviously I'm biased towards the solution I came to and other are biased to their own ideas.

I don't see biases as ill will I see them as human nature. I apologize if it read as ill will but it was not intended.

> As stated elsewhere no one has stated that this is a "bad PR" or "abhorrent" as you stated in the e-mail conversation. Again this is not really helpful.

This is the feeling I get from the conversation. I'm not stating facts as much as what I perceive from the conversation.

jgleissner:
For the record, I don't feel attached to the raw image uploader. In fact, it feels a bit odd to have two services that upload files, but the raw image uploader implementation fitted the use case "drop the image file into an S3 bucket" as an optional step in the image release process nicely enough.

The use case for the SAS upload changes things insofar that it introduces a new task for MASH, namely just uploading an image file, which is different from what MASH was designed to do. As I see it, we can either change the architecture of MASH to be able to model both processes nicely, or try to implement it without changing the architecture. The latter means either implementing it in the uploader or the raw-image-uploader (and possibly running it earlier in the pipeline), both options are not ideal IMHO. I'd prefer the redesign of the service pipeline as outlined by Sean.

smarlowucf:
The agreement here after further discussion was to merge the PR as is with the goal of getting consistency for uploads across cloud frameworks in the future.

PR #589, opened 2019-10-30
Fix retry region testing

### LINKED ISSUE #584: Testing service retry regions is broken
If retry is initiated in GCE testing service the service will crash due to key error. Region is used to drive and track results for jobs. When a random region is selected which is not part of the configured job no information for that region can be found. This functionality likely requires a fundamental redesign to fix the underlying bug and make it more maintainable.

smarlowucf:

- Simplify img-proof helper module
- Move service account file creation to job classes.
- Remove threading for ec2 testing
- Move exception handling to job classes (handled differently in gce)
- Return result from helper function instead of setting results in a dictionary
- Handle fallback regions in gce job class
- Move ec2 network setup to context manager

Unit & E2E tests

Fixes #584


mash/services/testing/gce_job.py
```diff
@@ -95,37 +96,55 @@ def run_job(self):
...
+                    )
+                except IpaRetryableError as error:
+                    result = {
+                        'status': EXCEPTION,
+                        'msg': str(error)
+                    }
+                    fallback_regions.remove(retry_region)
+
+                    if fallback_regions:
+                        retry_region = random.choice(fallback_regions)
```
rjschwei: You could drop this `if` condition and put the region randomization at the beginning of the while loop where it is guaranteed that `fallback_regions` is not an empty set. That would also eliminate line 112. Given that we have this retry loop I don't think it is important that we start testing in the "origin" region, i.e. `self.region`. The goal is really "find a region where we have no silly quota issues" and therefore where we start that "search" is somewhat immaterial.

But the code as implemented certainly also satisfies that functionality.

smarlowucf: I did consider this first, however came to the opposite conclusion. As a user I would be confused if I chose a certain target region and a test instance launched in some other region without any previous failure. It may lead to more likely failures. I.e. if I choose us-east1 to be my testing region and thus requested higher quota in only that region. When a job is started and MASH instead chooses some other region at random it's more likely to fail until it selects us-east1.

PR #598, opened 2019-11-15
Fix typo in job conditions schema.

### LINKED ISSUE #596: Build_id should be release
In image conditions build_id was renamed to release when obs-img-utils was ported out. This needs to be updated in mash schema and example job docs.

https://github.com/SUSE-Enceladus/mash/blob/master/mash/services/api/schema/jobs/__init__.py#L39

smarlowucf:

Build_id was renamed to release in obs-img-utils.

Unit tests.

### Additional Information

Closes #596


smarlowucf:
Requires fixes from #597 for passing build.

PR #599, opened 2019-11-15
Distro is an optional argument.

### LINKED ISSUE #593: Distro is optional
https://github.com/SUSE-Enceladus/mash/blob/master/mash/services/jobcreator/base_job.py#L39

smarlowucf:

Distro is an optional argument.

Unit tests.

### Additional Information

Fixes #593


smarlowucf:
Requires fixes from #597 for passing build.

PR #600, opened 2019-11-19
Add disk generation option for Azure publish.

smarlowucf:
## What does this PR do? Why are we making this change?

Publishing to a second generation is possible using disk generations by providing the generation_id and an optional suffix to be appended to the media name. If no suffix is provided the generation_id is appended instead.

Unit + integration. E2E when all changes merged and ready for deployment.

### DIFF
--- a/examples/messages/azure_job.json
+++ b/examples/messages/azure_job.json
@@ -29,7 +29,7 @@
   "publisher_id": "suse",
   "label": "New Image 123",
   "sku": "123",
-  "generation_id": "gen2",
+  "generation_id": "image-gen2",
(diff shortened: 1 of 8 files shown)


examples/messages/azure_job.json
```diff
@@ -29,6 +29,8 @@
   "publisher_id": "suse",
   "label": "New Image 123",
   "sku": "123",
+  "generation_id": "gen2",
+  "cloud_image_name_generation_suffix": "gen2",
```
bear454: Consider testing with different strings for `generation_id` and `cloud_image_name_generation_suffix`, since failover of one value to the other is part of the feature description.

PR #604, opened 2019-12-03
Add cloud_account option for ec2 jobs.

### LINKED ISSUE #595: EC2 job cloud account option
It would be helpful for there to be an additional cloud account option for ec2 jobs. This would match gce and azure and could be translated to cloud accounts format in processing. The new option would be xor with cloud_accounts otherwise an error should be raised.

smarlowucf:

A cloud_account name can be provided for the job instead of the cloud_accounts list or cloud_groups list.

Unit tests.

### Additional Information

Fixes #595

### DIFF
--- a/mash/services/api/schema/jobs/ec2.py
+++ b/mash/services/api/schema/jobs/ec2.py
@@ -54,6 +54,9 @@ ec2_job_message['properties']['cloud_accounts'] = {
     'minItems': 1,
     'example': [{'name': 'account1'}]
 }
+ec2_job_message['properties']['cloud_account'] = string_with_example(
+    'account1'
+)
 ec2_job_message['properties']['cloud_groups'] = {
     'type': 'array',
     'items': non_empty_string,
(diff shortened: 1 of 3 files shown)


mash/services/api/schema/jobs/ec2.py
```diff
@@ -62,6 +65,7 @@
     'example': ['group1']
 }
 ec2_job_message['anyOf'] = [
+    {'required': ['cloud_accounts']},
```
bear454: I think this is supposed to be `cloud_account` ?

rjschwei:
Appears superfluous. specifying one account is simply specifying a list of 1 and this case is covered by the existing `cloud_accounts` option.

smarlowucf:
> specifying one account is simply specifying a list of 1 and this case is covered by the existing `cloud_accounts` option.

Yes, it adds another route to perform the same function. However, it also provides consistency across all cloud frameworks and the `cloud_account` option is less confusing if a user is only using one account.

smarlowucf:
Going ahead with merge if there are no issues raised.

PR #609, opened 2019-12-20
Add upload consistency to all cloud frameworks

smarlowucf:

Make the workflow for raw image upload consistent across each cloud framework.

Inject job factory to listener class during service creation. Instead of the function being buried in the service it allows for smaller factories to be created based on the service type. This allows for configurable job classes and types. Supported job types are injected when creating a job factory and the factory is injected when creating a service.

Create a default no op job class and remove all duplicate no op job classes.

Create new create image service for handling framework image creation. Split GCE and Azure image creation steps from uploader service. Move EC2 uploader functions (lift/shift) to create service. Use NoOpJob class for EC2 upload which has no function.

Split out Azure SAS upload from Azure upload class. Now any image can be uploaded to S3 either as a one off job or as part of a publishing job. Also, Azure images can be uploaded via a SAS token as a one off upload or as part of a publishing job. Note: Azure images that are uploaded to S3 are not expanded as they are when uploaded to the Azue framework.

Unit and E2E tests.

- Update the configuration to add create service
- Notify admin to start tracking mash_create.service

### Additional Information

- Azure SAS arguments have been removed to make it more consistent with the S3 usage. To upload an image via Azure SAS the `raw_image_upload_type` can be set to `azure_sas` and the `raw_image_upload_location` should be a container SAS URL that follows the convention https://[storage-account].[maangement-url]/[container]?[SAS token].
- The data passed from service to service is more aggregated into a single arg (source_regions). This makes things more modular with a goal of paving the way to make the order of services more configurable. The naming (source_regions) is slightly confusing and thus should probably be changed in a future PR.
- Remove the process handling from Azure upload since the patch to fix the empty chunk method is added to the CI package of azure_storage_blob. The patch makes the upload I/O bound rather than CPU bound and thus removes the necessity for using processes.


mash/services/create/gce_job.py
```diff
@@ -0,0 +1,110 @@
...
+        self.cloud_image_description = format_string_with_date(
+            self.base_cloud_image_description,
+            timestamp=timestamp
+        )
+
+        self.request_credentials([self.account])
+        credentials = self.credentials[self.account]
+
+        with create_json_file(credentials) as auth_file:
+            object_name = ''.join([self.cloud_image_name, '.tar.gz'])
```
rjschwei: This name is already constructed on line 76 in uploader/gce_job.py to avoid breakage due to naming changes the name originally created by the uploader should be re-user here rather than re-constructing the name from scratch.

mash/services/uploader_service.py
```diff
@@ -33,13 +39,29 @@ def main():
...
+        service_name = 'uploader'
+
+        # Create job factory
+        job_factory = BaseJobFactory(
+            service_name=service_name,
+            job_types={
+                'azure': AzureUploaderJob,
+                'ec2': NoOpJob,
+                's3bucket': S3BucketUploaderJob,
+                'gce': GCEUploaderJob
```
rjschwei: Indicates the same problem as stated previously w.r.t. the use of the downloaded image file from the build service and indicates a misunderstanding in the requirements. We have to be able to upload any image file to an S3 bucket and then use that image file to create a VM image in any of the supported framework, image file type appropriate usage is expected. Meaning I can get an Azure image from the build service, upload it to an S3 bucket and use the same file to create a VM in Azure.

The current implementation would require the submission of 2 jobs to achieve this goal. However, with 2 jobs running it is possible to download one file for the s3 upload and another image file for the VM creation, this is dependent on the publishing cycle of the build service which we do not control.

mash/services/jobcreator/base_job.py
```diff
@@ -64,6 +64,14 @@ def __init__(self, kwargs):
...
         self.kwargs = kwargs
 
+        if not self.raw_image_upload_type:
+            self.raw_image_upload_type = self.cloud
+        elif self.raw_image_upload_type == 's3bucket' and self.last_service != 'uploader':
+            raise MashJobCreatorException(
+                'Cannot process image past uploader service '
+                'when uploading to an S3 bucket.'
+            )
+
```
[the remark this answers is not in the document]
smarlowucf: The ability to "sync" images in MASH is implemented as conditions which are handled in OBS service when a raw image is downloaded. These are not changing. The only change here is that with this PR you couldn't do a raw image upload in conjunction with framework image creation. However, the changes put EC2 workflow in a consistent state with Azure. Consistency was the biggest goal with these changes. If there is a need to test and or release images for multiple jobs then upload to an S3 bucket it now requires an extra job.

mash/services/replication_service.py
```diff
@@ -35,12 +40,25 @@ def main():
         log = logging.getLogger('MashService')
         log.setLevel(logging.INFO)
 
+        service_name = 'replication'
+
+        # Create job factory
+        job_factory = BaseJobFactory(
+            service_name=service_name,
+            job_types={
+                'azure': AzureReplicationJob,
```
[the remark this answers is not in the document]
smarlowucf: Azure has the extra step of copying the image from ARM to ASM. So you are correct that there is no replication but the job is implemented to handle this copying step. The step is supposed to go away in the future but there is no timetable for this change.

mash/services/jobcreator/base_job.py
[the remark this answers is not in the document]
rjschwei: This has nothing to do with "sync", it is the flow of "download image file -> put it into an S3 bucket -> use the same file for image creation" that is a required use case to satisfy the SAP CHOST image management that is being missed.

mash/services/uploader_service.py
smarlowucf: Right it requires 2 jobs but it's still possible by using conditions. We can guarantee that the same image build is used for both jobs. It's not as efficient and it's not the same file on disk but we can guarantee the builds are the same which should mean the packages are identical.

My goal with this PR was mainly to get to a consistent state for upload across cloud frameworks. And my understanding was that if we can lock an image at a build number this is sufficient for the use case?

jgleissner: I assume the build service will never use the same version twice, but downloading the file twice is still not ideal IMHO. One job could fail due to a network hiccup or whatever, and afterwards the file may already be gone from the build service, so you'd need to start all over again. Plus it seems unnecessary to download potentially large files twice.

Wouldn't it be possible to create multiple uploader jobs in one MASH job? I guess it would be a little fiddly because ListenerService isn't really designed to handle multiple jobs per job_id, but I suppose it could. Or perhaps the job creator could create a secondary MASH job with a different ID which really is just a second uploader job. For this to work the OBS service would need to send a second publish message with the secondary ID, but the OBS service is already special (i.e. not derived from ListenerService) so I guess it wouldn't be big deal, but I may be missing something.

smarlowucf:
@rjschwei @jgleissner With two upload services that both do raw image uploading it seems confusing to have one named upload and the other raw_image_upload. Are there any opinions against naming the second uploader post_upload to denote that it happens second in the pipeline?

obs -> upload -> testing -> post_upload -> replicate -> publish -> deprecate

Or maybe secondary_upload? Mainly I find raw_image_upload to be confusing now since the two services are essentially the same in terms of what functions they provide. The distinction is rather where they fall in the pipeline.

rjschwei:
@smarlowucf hmm, now I am confused. I thought that we had agreed not to upload anything after testing anymore. Thus resulting in

obs -> raw_img_upload -> upload -> testing ....

Where "raw_image_upload" sticks the file (.tar.gz, .raw.xz, .vhdfixed.xz) into either an S3 bucket or an Azure container.

W.r.t. using "upload" as a name twice, we could re-name the existing "upload" to "create". As part of create we also need to upload, which is a pre-condition but the service would then be named after the target primary functionality.

That was my mental picture based on our last conversation about the topic.

smarlowucf:
My understanding was we would first split create functions from upload service. Which I forgot in the the previous comment:

obs -> upload -> create -> testing -> post_upload -> replicate -> publish -> deprecate

This affects Azure and GCE since the image creation was combined in the upload classes previously. Then we would have a set of classes to handle raw uploads to; GCE bucket, Azure bucket, Azure SAS and S3 (EC2 image upload is part of creation and it's class is used by create service).

If upload is the last service then any 4 of those upload options is valid for the first and only upload. If last service is not upload then you cannot use S3 for the first upload service. For the second upload service S3 is the only option right now and it could be used by an image for any cloud. I.e. any job could upload the raw image tarball to S3 as part of a publishing job or after testing.

From what I pulled from all the conversations this was acceptable. The final piece if requested would be to enable Azure SAS uploads in the second upload service so they could be combined with a publishing job. This is where we should do some cleanup with job doc arguments so the options don't get too unwieldy from a user perspective.

As far as where the second upload sits in the pipeline there's no functional difference. Given this it seems better to have the second "raw image" upload after testing as it was before.

smarlowucf:
@rjschwei @jgleissner This is now ready for review again. I chose to go with the second option we had discussed previously. There are two uploader services that share upload classes. Any image can be uploaded to S3 as a one-off job or as part of a publishing job. Azure images have a second "raw" option to SAS URLs which has the same two options. The second "raw" image upload happens after testing as it exists already. This was all made possible by two main changes; splitting up upload and create into separate services and cleaning up the job factory to make it more modular.

I've updated the PR description with comments covering all changes. Note: prior to merge of this PR it should be squashed to clean up the git history.

mash/services/base_defaults.py
```diff
@@ -48,8 +48,8 @@ def get_log_directory(self):
     @classmethod
     def get_service_names(self):
         return [
-            'obs', 'uploader', 'testing', 'raw_image_uploader', 'replication',
-            'publisher', 'deprecation'
+            'obs', 'uploader', 'create', 'testing', 'raw_image_uploader',
+            'replication', 'publisher', 'deprecation'
```
rjschwei: Sorry, but I still do not understand the order. If I want `last_service: raw_image_uploader` which would be appropriate for a SAS URL upload only job then the has to flow through "uploader, create, and testing" services or these have to become special "NoOp" jobs.

What am I missing?

jgleissner: My understanding is that for an upload only job you would use `last_service: uploader` and create an `azure_sas` job. `raw_image_uploader` is really just a second instance of the uploader service.

rjschwei: Yes, which is why https://github.com/SUSE-Enceladus/mash/pull/609/files#diff-36578ceabc423a9513333c5858e0ad67R67 exists. But that would imply that we have to have a guard somewhere that does not allow `last_service: raw_image_uploader`, which I have not seen. Thanks for the reminder about that oddity.

smarlowucf: That is as intended. Last service being raw_image_uploader is another option a user would have. A way to upload a raw image but test it prior to upload. This is one use case we already had in place.

PR #620, opened 2020-02-05
Add configuration values for OCI operations.

smarlowucf:

### DIFF
--- a/mash/services/base_config.py
+++ b/mash/services/base_config.py
@@ -339,3 +339,34 @@ class BaseConfig(object):
         )
 
         return domain_whitelist or Defaults.get_domain_whitelist()
+
+    def get_max_oci_attempts(self):
+        """
+        Return the max number of attempts for OCI requests.
+
+        :return: int
+        """
+        max_oci_attempts = self._get_attribute(attribute='max_oci_attempts')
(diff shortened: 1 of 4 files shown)


rjschwei:
This code is fine, but the `get_oci_upload_process_count()` smells like a "global", i.e. across jobs limit function. I am not certain how we would "stall" jobs based on that limit. Let's say we send 4 OCI jobs into the pipeline, how would the fourth one know how to wait until one of the other 3 is done uploading. This violates our "jobs are independent" principal. But I am inferring this based on the name of the function and the tight scope of the submission.

smarlowucf:
> This code is fine, but the `get_oci_upload_process_count()` smells like a "global", i.e. across jobs limit function. I am not certain how we would "stall" jobs based on that limit. Let's say we send 4 OCI jobs into the pipeline, how would the fourth one know how to wait until one of the other 3 is done uploading. This violates our "jobs are independent" principal. But I am inferring this based on the name of the function and the tight scope of the submission.

This value is used by the SDK uploader method. It has an option for parallel upload across multiple processes. This is per image upload and not a global value. I see where the name could be confused but couldn't think of a better name that wasn't too long.

PR #627, opened 2020-02-17
Add OAuth2/OpenID Connect support

jgleissner:
This PR implements two new `auth` endpoints:

`/oauth2_req` returns an authorization URL to the client based on the MASH server configuration.
`/oauth2_login` performs user login based on authorization code the client retrieved from the authentication provider.

Manual testing has been done authenticating against `developer.okta.com`. Unit tests are missing at the moment.

Feature can be enabled in `/etc/mash_config.yaml`. This requires to configure the OAuth2 client secret in that file which is not supposed to be exposed, making that config file security sensitive.

### Additional Information

I've created the PR in a staging branch rather than master, as it's not quite complete yet (i.e. unit tests are missing). I wanted to discuss first whether the general approach is fine.


jgleissner:
Client side implemention: https://github.com/SUSE-Enceladus/mash-client/pull/52

mash/services/api/config.py
```diff
@@ -75,3 +75,23 @@ def EMAIL_WHITELIST(self):
...
+    @property
+    def OAUTH2_CLIENT_SECRET(self):
+        return self.config.get_oauth2_client_secret()
+
+    @property
+    def OAUTH2_PROVIDER_URL(self):
+        return self.config.get_oauth2_provider_url()
+
+    @property
+    def OAUTH2_REDIRECT_PORT(self):
```
smarlowucf: I think this is really better as a mash client decision. Details below.

mash/services/api/routes/auth.py
```diff
@@ -117,3 +150,107 @@ def delete(self):
...
+        """
+        Request oauth2 login URL.
+        """
+        data = json.loads(request.data.decode())
+        username = data['username']
+
+        if 'oauth2' not in current_app.config['AUTH_METHODS']:
+            return make_response(jsonify({'msg': 'OAuth2 login is disabled'}), 403)
+
+        if not get_user_by_username(username):
```
smarlowucf: Although I see the advantage here I don't think this is really possible. A user should be able to login into mash from a SSO (oidc provider) without previously creating an account.

mash/services/api/routes/auth.py
```diff
@@ -117,3 +150,107 @@ def delete(self):
...
+            return make_response(jsonify({'msg': 'Username is invalid'}), 401)
+
+        oauth2_auth_url = '{}/authorize'.format(
+            current_app.config['OAUTH2_PROVIDER_URL'],
+        )
+        oauth2_redirect_url = 'http://localhost:{}'.format(
```
smarlowucf: I think this should really be a decision made by the client. If it's made by the server then the client can end up in a dead end error if the user's machine happens to be using the port chosen by the server.

Ideally the client would have a range of ports that it could iterate through until it found one that was free. Then it would create the session based on that. Which I think would mean that the server sends back the info as json only.

I.e. the client_id, oauth URL, and scope.

mash/services/api/routes/auth.py
```diff
@@ -117,3 +150,107 @@ def delete(self):
...
+@api.route('/oauth2_login')
+class OAuth2Login(Resource):
+    @api.doc('oauth2_login')
+    @api.expect(oauth2_login_request)
+    @api.response(200, 'Logged in', oauth2_login_response)
+    @api.response(401, 'Unauthorized', default_response)
+    @api.response(403, 'Forbidden', default_response)
+    def post(self):
+        data = json.loads(request.data.decode())
+        username = data['username']
```
smarlowucf: I don't think the username is really necessary. The validation of a user's identity is coming from the identity provider. This should contain the username and/or email. And the username shouldn't be required if it's the users first login.

mash/services/api/routes/auth.py
```diff
@@ -117,3 +150,107 @@ def delete(self):
...
+        auth_code = data['auth_code']
+        state = data['state']
+
+        if 'oauth2' not in current_app.config['AUTH_METHODS']:
+            return make_response(jsonify({'msg': 'OAuth2 login is disabled'}), 403)
+
+        oauth2_redirect_url = 'http://localhost:{}'.format(
```
smarlowucf: Is the redirect URL required for this authorization request? I thought it was a synchronous request where the identity provider responds with the tokens directly to the request?

mash/services/api/routes/auth.py
```diff
@@ -117,3 +150,107 @@ def delete(self):
...
+        )
+        token = oauth2.fetch_token(
+            oauth2_token_url,
+            client_secret=oauth2_client_secret,
+            code=auth_code
+        )
+
+        # FIXME: verify signature
+        user_email = jwt.decode(token['id_token'], verify=False)['email']
+        if verify_email(username, user_email):
```
smarlowucf: I think this would be a `does_user_exist` check which would have two routes:

- If the user exists get tokens for that user.
- If the user does not exist yet then the user is created and we would need to generate a long random password for the user. Then reply back with the tokens.

mash/services/api/routes/auth.py
jgleissner: The thing is, the redirect URI must match what is configured in the OAuth provider. The client can't just pick a random port. Hence the idea to supply that information from the server, so you'd only have to configure it once and, assuming the server is set up correctly, is guaranteed to work. I see the issue with the port being taken already. We could support a list of port numbers, as Okta (and probably other OpenID Connect providers) allows to specific several redirect URIs. So you could offer the client one or even several fallback ports.

mash/services/api/routes/auth.py
jgleissner: It is synchronous but it requires the URL and verifies it.

mash/services/api/routes/auth.py
smarlowucf: Yeah, from Erik's comment it sounds like you can configured any number of URI's. So while it might not be fun if it has to be submitted manually we should be able to add many ports right?

Such as adding all URI's from localhost:9000 - localhost:9100 for example or maybe a smaller set. But having some options I think would be nice.

The thing is that if a user ran into issues with a port being used I think it's going to be very confusing to them as to why a server is running in the first place. While this could and probably should be documented alongside the client code I think it should really be out-of-sight/out-of-mind from a user perspective.

mash/services/api/routes/auth.py
smarlowucf: Ack. If the client chose the URI I think it's easy enough for the client to provide this info to the server in the login request.

mash/services/api/routes/auth.py
smarlowucf: One thing to consider. If there's no username coming from Okta or the username is the email address. Then we could drop username from Mash. And instead differentiate users via email. This would require more changes to the data model but if the extra data is not needed it simplifies Mash authentication.

mash/services/api/routes/auth.py
jgleissner: I'm not sure how many ports are supported, but I think it's unreasonable to expect any admin to configure 100 ports for one application. I also think it's already pretty unlikely that even two ports picked out of a relatively seldomly used range will cause a conflict on a typical user machine.

mash/services/api/routes/auth.py
jgleissner: The username in Okta appears to be the email address. You can additionally request a user profile from Okta, which provides you with the user's name (the actual name) and a `preferred_username` field, but that just seems to be the same as email address.

I guess the question is, do we need to bother at all with user names? I thought we may want to couple oidc users with mash users for access control purposes, but thinking about it, that doesn't make too much sense. Any user who matches the email whitelist can create an account. So I suppose we might as well just check the email address we get from Okta against the whitelist, and drop  theusername from the oidc auth flow entirely.

smarlowucf: My opinion would be to drop username as a whole from mash. This would have a lot of trickle down changes as email would have to replace username in a lot of places.

Also, I think that a user is the same no matter what route they login. I.e. if I have an account already with an email test@test.com which was created using the user/password flow. Then I log in via a OIDC auth flow which returns the same email test@test.com then the account that already exists is returned. The user object is always the same if the email matches they just have multiple routes to authenticate now.

smarlowucf: My opinion would be to drop username as a whole from mash. This would have a lot of trickle down changes as email would have to replace username in a lot of places.

Also, I think that a user is the same no matter what route they login. I.e. if I have an account already with an email test@test.com which was created using the user/password flow. Then I log in via a OIDC auth flow which returns the same email test@test.com then the account that already exists is returned. The user object is always the same if the email matches they just have multiple routes to authenticate now.

mash/services/api/routes/auth.py
smarlowucf: That's fair 100 or even 25 is probably more than necessary. I still think having at least a couple available options which the client can roll through is a good solution. Having one port and the server choose that port means the only way forward for a user is to unblock that port.

My suggestion would be have at least a few options and consult something like https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=113 to find the best range of unreserved or unlikely to be used ports.

smarlowucf: As I think about this more I realize that since the server is handing out information already about the auth provider then handing out what redirect URIs to use is okay. Maybe we can use an upper and lower bound or a range option? So the values and number of ports can be configured like: `oauth_redirect_ports=9000:9005` or

```
oauth_redirect_starting_port=9000
oauth_redirect_port_range=5
```

jgleissner: I've implemented it as a list rather than a range. I think it's more useful as it gives you the flexibility to use completely different ports, and I don't think a large number of ports make a lot of sense.

mash/services/api/routes/auth.py
jgleissner: I've updated the implementation to drop the username from the OIDC auth workflow. The email address provided by the auth provider is used to check against the whitelist. In case a user authenticates via OIDC for the first time, and there's no account associated with the email address, an account is created with a random password (effectively making it a non-password based account). If there is already an account with that email, that one will be used, so you could switch to OIDC with existing users intact.

The implementation uses email as username. I guess we could get rid of username altogether and just use the email as identification. As you said, it would require more changes, with no tangible benefit, but then username is a bit redundant in that setup.

mash/services/api/routes/auth.py
smarlowucf: Sounds good. :+1:

mash/services/api/routes/auth.py
smarlowucf: It will actually add a decent bit of complication if we had both routes where a user could have an actual username or only an email because this has to be handled every time credentials are added, retrieved, deleted etc. So I think the benefit is simplifying things. The reason we went with having a username was because the user otherwise had to type in their email every login. If our focus is on authentication via OIDC then this is no longer an issue. Plus the fact that we are using JWT tokens means that you only login once to get a long running session and the client automagically handles token refresh until that session is expired.

jgleissner: I'm a bit uncertain what the verdict is. If we want to keep password based authentication, I guess we will also want to keep the username, right? In that case the current implementation would do. The automatically created accounts simply use email for both the username and the email fields.

If we want to simplify and get rid of the username, we could basically drop the email field and use email addresses for the username field, which I believe wouldn't require a lot of code changes. Password based authentication would still be possible but with somewhat awkward usernames.

We could also remove password based authentication, but that would make it harder to set up a local MASH server (e.g. for testing).

mash/services/api/routes/auth.py
```diff
@@ -117,3 +148,92 @@ def delete(self):
...
+
+    @api.expect(oauth2_login_request)
+    @api.response(200, 'Logged in', oauth2_login_response)
+    def post(self):
+        data = json.loads(request.data.decode())
+        auth_code = data['auth_code']
+        state = data['state']
+        redirect_port = data['redirect_port']
+
+        if 'oauth2' not in current_app.config['AUTH_METHODS']:
```
rjschwei: Test should be first.

mash/services/api/routes/auth.py
rjschwei: Yes, we want both. It was always agreed that the authentication backend would be "plugable" and we implement what we need, password based initially and now openid with Okta.

smarlowucf: I think it's okay to keep password authentication but we need to be careful to ensure that each route is equivalent and seamless.

If we keep username this becomes difficult because of the different permutations a user could choose. For example if a user creates an account via SSO they have an email but no username and a randomly generated password. Whereas if they create an account via password auth first they have a username, email and chosen password.

If they go "password" -> "okta" then things are seamless. But the other way around is not since it currently requires username but the user wouldn't have one.

My suggestion is that we drop username to make things more seamless. If there's concern about password auth requiring email address I think this can be resolved by having a client config option to set email once so it's optional during login:

``` bash
> mash auth login
"Note: Logging in as test@test.com"
Enter password: 
Confirm password:
"Logged in"
```

Where the user previously set the email to use in client config:

``` yaml
email: test@test.com
```

In either case I think we need to drop usage of username or email as a primary key used for credential storage. I think we need to have a follow on pull request that changes the usage of username to be user_id.

rjschwei: > If we keep username this becomes difficult because of the different permutations a user could choose. For example if a user creates an account via SSO they have an email but no username and a randomly generated password. Whereas if they create an account via password auth first they have a username, email and chosen password.
> 

That should not happen. Once the server gets swicthed to use SSO any local authentication becomes invalid, i.e. all accounts that exists in the DB today will be void. We can decide whether the DB should be cleaned or continue to exists in case we swicth back to local auth. But there cannot be intermixing. Only one auth method is active.


> In either case I think we need to drop usage of username or email as a primary key used for credential storage. I think we need to have a follow on pull request that changes the usage of username to be user_id.

Hmm, `user_id` as in a generated id? That would work for me.

smarlowucf: Currently AUTH_METHODS is a list which would allow for both routes to be active at once. I think either way is possible (single/multiple auth options) but if we want only one then the config should probably be AUTH_METHOD as a string. If we go this route then username could be changed to optional that way it can be ignored by SSO auth.

Yes, user_id that is generated as PK in database for a user. I originally chose the username for creds service to make the dir structure look nice but this was not the best functional decision.

jgleissner: It's easy enough to change it to allow only one method. The email address is currently simply used as username as well, but I guess we could mark it as optional in the database so we wouldn't need to set one.

Wiping the user database should not really be necessary when going from password to oauth, but when switching from oauth to password the automatically created accounts would need to be purged. We could add a simple tool that does that. I don't think we'd need to go any more fancy as that scenario is probably rather rare.

smarlowucf: The migration is where dropping username would be advantageous. I haven't done so with Alembic yet but I imagine we could create data migrations if need be based on DB changes. The difficulty is handling credentials migration right now but this would be resolved by moving to user_id.

Agree, that it's not likely this would change much in production to be a concern but if the transition were available in a smooth fashion that would be advantageous.

jgleissner:
I've added unit tests, so hopefully this PR is complete now. I think it might be easier to handle to do the username change in a separate PR.

mash/services/api/routes/auth.py
```diff
@@ -117,3 +146,90 @@ def delete(self):
...
+            scope="openid email",
+            state=state
+        )
+        token = oauth2.fetch_token(
+            oauth2_token_url,
+            client_secret=oauth2_client_secret,
+            code=auth_code
+        )
+
+        # FIXME: verify signature
```
smarlowucf: Todo?

smarlowucf: Todo?

mash/services/api/routes/auth.py
```diff
@@ -117,3 +146,90 @@ def delete(self):
                 jsonify({'msg': 'Logout failed'}),
                 400
             )
+
+
+@api.route('/oauth2')
+@api.doc('oauth2')
+@api.response(401, 'Unauthorized', default_response)
+@api.response(403, 'Forbidden', default_response)
+class OAuth2Req(Resource):
```
smarlowucf: Suggestion: Expand class name to OAuth2Request? But seems like a confusing class name for a route. I cannot think of a different name off the top of my head. :thinking:

mash/services/api/routes/auth.py
```diff
@@ -117,3 +146,90 @@ def delete(self):
...
+
+            response = {
+                'access_token': access_token,
+                'refresh_token': refresh_token
+            }
+            return make_response(jsonify(response), 200)
+        else:
+            current_app.logger.warning(
+                'Failed login attempt for user: {username}'.format(
+                    username=user_email
```
smarlowucf: Nitpick: /username/email or user_email.

mash/services/api/utils/users.py
```diff
@@ -51,18 +48,38 @@ def add_user(username, email, password):
         username=username,
         email=email
     )
+    if not password:
+        password = ''.join(
+            [choice(string.ascii_letters + string.digits) for n in range(64)]
```
smarlowucf: This is probably okay but it is recommended to use random from secrets module for secure data. This would lock mash at python3 >= 3.6 which I think is okay for us as Mash is running in SLE15 which has 3.6.

https://docs.python.org/3/library/secrets.html#module-secrets

smarlowucf:
Using secrets I think is good but we'll need to drop 3.5 from testing, update the list in setup.py and probably would be good to mark the min python requirement in spec.

jgleissner:
> Using secrets I think is good but we'll need to drop 3.5 from testing, update the list in setup.py and probably would be good to mark the min python requirement in spec. 3.5 is nearing it's EOL anyway.

Alternatively I we could fall back to the implementation in `random` if `secrets` is not available. It would be dead simple as usage is the same. But if 3.5 is nearly EOL we might as well drop support.

smarlowucf:
> Alternatively I we could fall back to the implementation in `random` if `secrets` is not available. It would be dead simple as usage is the same. But if 3.5 is nearly EOL we might as well drop support.

EOL is half a year out (2020-09-13). I think either solution is okay. For our use case we are on 3.6 so there's no harm in dropping 3.5 support.

jgleissner:
I've changed authentication method parameter back to a list to allow password based accounts alongside oauth2 for special accounts. I suppose another review of the latest commits might be useful before merge. I'm working on removing username in another branch. I'll rebase it onto `oauth-staging` and open a PR once this is merged.

smarlowucf:
Looks like the last piece is the python 3.5 compatibility. Either to drop support or add a conditional import for secrets.

jgleissner:
> Looks like the last piece is the python 3.5 compatibility. Either to drop support or add a conditional import for secrets.

I removed the line with Python 3.5 from setup.py. Is there anything more I need to do?

smarlowucf:
3.5 testing can be dropped from Travis to prevent failure in checks. See https://github.com/SUSE-Enceladus/mash/blob/master/.travis.yml

jgleissner:
> 3.5 testing can be dropped from Travis to prevent failure in checks. See https://github.com/SUSE-Enceladus/mash/blob/master/.travis.yml

Thanks, removed.

PR #640, opened 2020-03-20
Merge OAuth2/OIDC implementation into master

jgleissner:
PR to merge the OAUth2 OpenID Connect support implementation into master.

One thing, the configuration parameters this adds to the MASH server aren't reflected in the example configuration. Should we add them in some way or form?


smarlowucf:
> One thing, the configuration parameters this adds to the MASH server aren't reflected in the example configuration. Should we add them in some way or form?

I don't know if I've added every param to the example config. At least the required ones but it doesn't hurt.

PR #646, opened 2020-04-10
Update descriptions for all job doc args.

smarlowucf:

Update descriptions for all job doc args.

### DIFF
--- a/mash/services/api/schema/jobs/__init__.py
+++ b/mash/services/api/schema/jobs/__init__.py
@@ -44,7 +44,12 @@ image_conditions = {
 
 utctime = {
     'type': 'string',
-    'description': 'An RFC3339 date-time string, "now" or "always"',
+    'description': 'An RFC3339 date-time string, "now" or "always".'
+                   'If using a date string it must be in the future '
+                   'and the job will start no sooner than the provided '
+                   'date. Now jobs will run as soon as possible and always '
+                   'jobs run through testing every time a new image tarball '
+                   'is published.',
     'format': 'regex',
     'pattern': r'^([0-9]+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]'
                r'|3[01])[Tt]([01][0-9]|2[0-3]):([0-5][0-9]):'
@@ -68,7 +73,9 @@ base_job_message = {
                 'publisher',
                 'deprecation'
             ],
-            'example': 'create'
+            'example': 'create',
+            'description': 'Where the job should finish. If only uploading '
+                           'an image then uploader would be the last service.'
         },
         'utctime': utctime,
         'image': string_with_example('openSUSE-Leap-15.0-EC2-HVM'),
@@ -88,7 +95,15 @@ base_job_message = {
                     'version': '4.13.1',
                     'condition': '>='
                 }
-            ]
+            ],
+            'description': 'A list of image conditions to check the image '
+                           'against. Providing only the package name will '
+                           'ensure the package is in the image. The version '
+                           'is the package version from the build service '
+                           'whereas the release is the Kiwi build number. '
+                           'At least one of package_name, version or release '
+                           'is required. If no package name is provided then '
+                           'the condition is against the image itself.'
         },
         'download_url': string_with_example(
             'https://download.opensuse.org/repositories/'
@@ -100,16 +115,28 @@ base_job_message = {
         'distro': {
             'type': 'string',
             'enum': ['opensuse_leap', 'sles'],
-            'example': 'sles'
+            'example': 'sles',
+            'description': 'The distro is used for img-proof testing.'
         },
         'instance_type': string_with_example('t2.micro'),
         'tests': {
             'type': 'array',
             'items': non_empty_string,
             'minItems': 1,
-            'example': ['test_sles']
+            'example': ['test_sles'],
+            'description': 'This is a list of img-proof tests or test '
+                           'descriptions. The tests will be run against '
+                           'the instance. If any tests fail the job will '
+                           'fail.'
         },
-        'cleanup_images': {'type': 'boolean', 'example': True},
+        'cleanup_images': {
+            'type': 'boolean',
+            'example': True,
+            'description': 'Whether to cleanup the image artifacts. By '
+                           'default all artifacts are cleaned up when the '
+                           'last service is testing or if a publishing job '
+                           'fails.'
+        },
         'cloud_architecture': {
             'type': 'string',
             'enum': ['x86_64', 'aarch64'],
@@ -119,15 +146,19 @@ base_job_message = {
         'notification_type': {
             'type': 'string',
             'enum': ['periodic', 'single'],
-            'example': 'single'
+            'example': 'single',
+            'description': 'The single notification option sends an email '
+                           'with job status when the job finishes or fails. '
+                           'The periodic option will send an email after the '
+                           'finishes each service with the job status.'
         },
         'profile': string_with_example('Proxy'),
         'conditions_wait_time': {
             'type': 'integer',
             'minimum': 0,
             'example': 900,
-            'description': 'Time (in seconds) to wait before failing '
-                           'on image conditions.'
+            'description': 'The time (in seconds) to wait before failing '
+                           'a job on image conditions.'
         },
         'raw_image_upload_type': string_with_example('s3bucket'),
         'raw_image_upload_location': string_with_example('my-bucket/prefix/'),
@@ -136,13 +167,20 @@ base_job_message = {
             'type': 'array',
             'items': non_empty_string,
             'minItems': 1,
-            'example': ['MIT']
+            'example': ['MIT'],
+            'description': 'A list of licenses that should not be in the '
+                           'image. If a package is in the manifest that has '
+                           'a matching license the job will fail.'
         },
         'disallow_packages': {
             'type': 'array',
             'items': non_empty_string,
             'minItems': 1,
-            'example': ['*-mini']
+            'example': ['*-mini'],
+            'description': 'A list of packages that should not be in the '
+                           'image. If a package is in the manifest that has '
+                           'a matching name the job will fail. The * wildcard '
+                           'can be used to match a package name pattern.'
         },
     },
     'additionalProperties': False,
--- a/mash/services/api/schema/jobs/__init__.py
+++ b/mash/services/api/schema/jobs/__init__.py
@@ -25,13 +25,26 @@ from mash.services.api.schema import (
 image_conditions = {
     'type': 'object',
     'properties': {
-        'package_name': string_with_example('kernel-default'),
-        'version': string_with_example('4.13.1'),
-        'release': string_with_example('1.1'),
+        'package_name': string_with_example(
+            'kernel-default',
+            description='The name of the package for this condition. '
+                        'If no name is provided for a condition then the '
+                        'condition is checked against the image itself.'
+        ),
+        'version': string_with_example(
+            '4.13.1',
+            description='The package version from the build service.'
+        ),
+        'release': string_with_example(
+            '1.1',
+            description='The Kiwi build (release) number.'
+        ),
         'condition': {
             'type': 'string',
             'enum': ['>=', '==', '<=', '>', '<'],
-            'example': '=='
+            'example': '==',
+            'description': 'The expression to use for comparing version '
+                           'and/or release.'
         }
     },
     'additionalProperties': False,
@@ -78,12 +91,27 @@ base_job_message = {
                            'an image then uploader would be the last service.'
         },
         'utctime': utctime,
-        'image': string_with_example('openSUSE-Leap-15.0-EC2-HVM'),
+        'image': string_with_example(
+            'openSUSE-Leap-15.0-EC2-HVM',
+            description='This should match the name of the tarball file on'
+                        'the download server prior to the architecture. For '
+                        'a file like openSUSE-Leap-15.0-EC2-HVM.x86_64-1.0.0-'
+                        'Build1.206.vhdfixed.xz.sha256 The "image" is '
+                        'openSUSE-Leap-15.0-EC2-HVM.'
+        ),
         'cloud_image_name': string_with_example(
-            'openSUSE-Leap-15.0-v{date}-hvm-ssd-x86_64'
+            'openSUSE-Leap-15.0-v{date}-hvm-ssd-x86_64',
+            description='The name to use for the uploaded image in the cloud '
+                        'framework. The name can have a date of upload '
+                        'inserted such as the example. The {date} will be '
+                        'replaced by the current iso date format at time of '
+                        'upload (20200410).'
         ),
         'old_cloud_image_name': string_with_example(
-            'openSUSE-Leap-15.0-v20190313-hvm-ssd-x86_64'
+            'openSUSE-Leap-15.0-v20190313-hvm-ssd-x86_64',
+            description='The cloud image name for the image to be deprecated. '
+                        'This is only required for jobs that run through '
+                        'deprecation service.'
         ),
         'conditions': {
             'type': 'array',
@@ -107,10 +135,15 @@ base_job_message = {
         },
         'download_url': string_with_example(
             'https://download.opensuse.org/repositories/'
-            'Cloud:/Images:/Leap_15.0/images/'
+            'Cloud:/Images:/Leap_15.0/images/',
+            description='The URL to a download repository. The URL is '
+                        'expected to have the image tarball, checksum and '
+                        'a packages file.'
         ),
         'image_description': string_with_example(
-            'openSUSE Leap 15.0 (HVM, 64-bit, SSD-Backed)'
+            'openSUSE Leap 15.0 (HVM, 64-bit, SSD-Backed)',
+            description='Description to use when creating the image in the '
+                        'cloud framework.'
         ),
         'distro': {
             'type': 'string',
@@ -118,7 +151,11 @@ base_job_message = {
             'example': 'sles',
             'description': 'The distro is used for img-proof testing.'
         },
-        'instance_type': string_with_example('t2.micro'),
+        'instance_type': string_with_example(
+            't2.micro',
+            description='Instance size/type img-proof will use when '
+                        'launching a test instance.'
+        ),
         'tests': {
             'type': 'array',
             'items': non_empty_string,
@@ -140,7 +177,10 @@ base_job_message = {
         'cloud_architecture': {
             'type': 'string',
             'enum': ['x86_64', 'aarch64'],
-            'example': 'x86_64'
+            'example': 'x86_64',
+            'description': 'The underlying architecture for the image. '
+                           'Valid options are x86_64 and aarch64 and the '
+                           'default architecture is x86_64.'
         },
         'notification_email': email,
         'notification_type': {
@@ -160,9 +200,22 @@ base_job_message = {
             'description': 'The time (in seconds) to wait before failing '
                            'a job on image conditions.'
         },
-        'raw_image_upload_type': string_with_example('s3bucket'),
-        'raw_image_upload_location': string_with_example('my-bucket/prefix/'),
-        'raw_image_upload_account': string_with_example('my_aws_account'),
+        'raw_image_upload_type': string_with_example(
+            's3bucket',
+            description='Cloud framework to use for raw image tarball '
+                        'upload. The image tarball will be uploaded as is '
+                        'to the framework using the cloud image name.'
+        ),
+        'raw_image_upload_location': string_with_example(
+            'my-bucket/prefix/',
+            description='The location to put the raw image. This is '
+                        'dependent on the cloud framework chosen.'
+        ),
+        'raw_image_upload_account': string_with_example(
+            'my_aws_account',
+            description='The account credentials to use for uploading the '
+                        'raw image.'
+        ),
         'disallow_licenses': {
             'type': 'array',
             'items': non_empty_string,
--- a/mash/services/api/schema/jobs/__init__.py
+++ b/mash/services/api/schema/jobs/__init__.py
@@ -37,7 +37,7 @@ image_conditions = {
         ),
         'release': string_with_example(
             '1.1',
-            description='The Kiwi build (release) number.'
+            description='The build (release) number.'
         ),
         'condition': {
             'type': 'string',
@@ -190,7 +190,7 @@ base_job_message = {
             'description': 'The single notification option sends an email '
                            'with job status when the job finishes or fails. '
                            'The periodic option will send an email after the '
-                           'finishes each service with the job status.'
+                           'job finishes each service with the job status.'
         },
         'profile': string_with_example('Proxy'),
         'conditions_wait_time': {
--- a/mash/services/api/schema/jobs/__init__.py
+++ b/mash/services/api/schema/jobs/__init__.py
@@ -33,11 +33,15 @@ image_conditions = {
         ),
         'version': string_with_example(
             '4.13.1',
-            description='The package version from the build service.'
+            description='The package or image version from the build service.'
+                        ' If no package_name is provided with the condition '
+                        'then the condition is against the image.'
         ),
         'release': string_with_example(
             '1.1',
-            description='The build (release) number.'
+            description='The build (release) number for the package or image.'
+                        ' If no package_name is provided with the condition '
+                        'then the condition is against the image.'
         ),
         'condition': {
             'type': 'string',
@@ -61,8 +65,8 @@ utctime = {
                    'If using a date string it must be in the future '
                    'and the job will start no sooner than the provided '
                    'date. Now jobs will run as soon as possible and always '
-                   'jobs run through testing every time a new image tarball '
-                   'is published.',
+                   'jobs run through the pipeline every time a new image '
+                   'tarball is published.',
     'format': 'regex',
     'pattern': r'^([0-9]+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]'
                r'|3[01])[Tt]([01][0-9]|2[0-3]):([0-5][0-9]):'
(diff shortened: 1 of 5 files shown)


mash/services/api/schema/jobs/__init__.py
```diff
@@ -25,13 +25,26 @@
...
+                        'condition is checked against the image itself.'
+        ),
+        'version': string_with_example(
+            '4.13.1',
+            description='The package version from the build service.'
+        ),
+        'release': string_with_example(
+            '1.1',
+            description='The Kiwi build (release) number.'
+        ),
```
jgleissner, line 41: Possibly nit-picking, but the release number is set by the build service. If anything, it's the version number that is set in the kiwi description, but that only applies to image builds, of course. Maybe just drop `Kiwi`.

mash/services/api/schema/jobs/__init__.py
```diff
@@ -88,61 +123,117 @@
...
         'notification_email': email,
         'notification_type': {
             'type': 'string',
             'enum': ['periodic', 'single'],
-            'example': 'single'
+            'example': 'single',
+            'description': 'The single notification option sends an email '
+                           'with job status when the job finishes or fails. '
+                           'The periodic option will send an email after the '
+                           'finishes each service with the job status.'
```
jgleissner, line 193: The last sentence lacks a subject (`job` perhaps?).

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
 
 azure_job_message = copy.deepcopy(base_job_message)
 azure_job_message['properties']['emails'] = string_with_example(
-    'test@fake.com'
+    'test@fake.com',
+    description='A Comma-separated list of email addresses to be notified '
```
jgleissner, line 27: I don't think comma is an important enough word to be capitalized. ;)

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
+    description='The title to be displayed in the marketplace.'
+)
+azure_job_message['properties']['offer_id'] = string_with_example(
+    'leap',
+    description='The name of a group of related images created by a '
+                'publisher.'
+)
+azure_job_message['properties']['publisher_id'] = string_with_example(
+    'suse',
+    description='The organization that created the image'
```
jgleissner, line 41: Maybe add a period for consistency.

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
+                'multiple disk generations.'
 )
-azure_job_message['properties']['offer_id'] = string_with_example('leap')
-azure_job_message['properties']['publisher_id'] = string_with_example('suse')
-azure_job_message['properties']['sku'] = string_with_example('15')
-azure_job_message['properties']['generation_id'] = string_with_example('gen2')
 azure_job_message['properties']['cloud_image_name_generation_suffix'] = string_with_example(
-    'gen2'
+    'gen2',
+    description='The suffix is appended to the cloud image name '
```
jgleissner, line 55: Ditto.

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
     'format': 'regex',
     'pattern': '^[0-9]{12}(,[0-9]{12})*$|^(all|none)$',
     'example': '123456789012,098765432109',
-    'examples': ['all', 'none', '123456789012,098765432109']
+    'examples': ['all', 'none', '123456789012,098765432109'],
+    'description': 'Sharing image to all shares the image publicly. '
+                   'Sharing to none keeps the image private and sharing '
+                   'to a comma separated list of accounts makes it available '
```
jgleissner, line 32: For consistency reasons, I'd either put a hyphen between comma and separated or remove it in mash/services/api/schema/jobs/azure.py. FWIW, I'm pro hyphen in this case. ;)

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
 ec2_job_message['properties']['share_with'] = share_with
-ec2_job_message['properties']['allow_copy'] = {'type': 'boolean'}
+ec2_job_message['properties']['allow_copy'] = {
+    'type': 'boolean',
+    'description': 'Whether to allow copy of the image.'
+}
 ec2_job_message['properties']['billing_codes'] = string_with_example(
-    'bp-1234567890,bp-0987654321'
+    'bp-1234567890,bp-0987654321',
+    description='A comma separated list of billing codes to apply during '
```
jgleissner, line 75: Ditto.

mash/services/api/schema/jobs/__init__.py
```diff
@@ -44,7 +57,12 @@
 
 utctime = {
     'type': 'string',
-    'description': 'An RFC3339 date-time string, "now" or "always"',
+    'description': 'An RFC3339 date-time string, "now" or "always".'
+                   'If using a date string it must be in the future '
+                   'and the job will start no sooner than the provided '
+                   'date. Now jobs will run as soon as possible and always '
+                   'jobs run through testing every time a new image tarball '
+                   'is published.',
```
rjschwei, line 65: s/testing/the pipeline/

There is nothing in mash to prevent a "publish every image build" approach.

mash/services/api/schema/jobs/__init__.py
```diff
@@ -68,15 +86,32 @@
                 'publisher',
                 'deprecation'
             ],
-            'example': 'create'
+            'example': 'create',
+            'description': 'Where the job should finish. If only uploading '
+                           'an image then uploader would be the last service.'
```
rjschwei, line 91: Suggest:
"The last service in the pipeline to be executed. Possible values are: upload -> Upload the image file to an Azure containre or S3 bucket; create -> Create an image; testing -> test the image......."

mash/services/api/schema/jobs/__init__.py
```diff
@@ -88,61 +123,117 @@
...
-            ]
+            ],
+            'description': 'A list of image conditions to check the image '
+                           'against. Providing only the package name will '
+                           'ensure the package is in the image. The version '
+                           'is the package version from the build service '
+                           'whereas the release is the Kiwi build number. '
+                           'At least one of package_name, version or release '
+                           'is required. If no package name is provided then '
+                           'the condition is against the image itself.'
```
rjschwei, line 134: Also explain the available condition operators. Is it "=" or "==" for example?

mash/services/api/schema/jobs/__init__.py
```diff
@@ -88,61 +123,117 @@
...
+            'openSUSE Leap 15.0 (HVM, 64-bit, SSD-Backed)',
+            description='Description to use when creating the image in the '
+                        'cloud framework.'
         ),
         'distro': {
             'type': 'string',
             'enum': ['opensuse_leap', 'sles'],
-            'example': 'sles'
+            'example': 'sles',
+            'description': 'The distro is used for img-proof testing.'
```
rjschwei, line 152: Suggest:
"The distribution setting used for the img-proof tests"

mash/services/api/schema/jobs/__init__.py
```diff
@@ -88,61 +123,117 @@
...
         },
-        'instance_type': string_with_example('t2.micro'),
+        'instance_type': string_with_example(
+            't2.micro',
+            description='Instance size/type img-proof will use when '
+                        'launching a test instance.'
```
rjschwei, line 157: Do we not have a condition where img-proof will select a random instance type from a pre-configured list? I thought we did. In that case we want to explain in the description that the value is optional and the pre-configured list wil be used to select a test instance.

mash/services/api/schema/jobs/__init__.py
```diff
@@ -88,61 +123,117 @@
...
+        'raw_image_upload_location': string_with_example(
+            'my-bucket/prefix/',
+            description='The location to put the raw image. This is '
+                        'dependent on the cloud framework chosen.'
+        ),
+        'raw_image_upload_account': string_with_example(
+            'my_aws_account',
+            description='The account credentials to use for uploading the '
+                        'raw image.'
+        ),
```
rjschwei, line 218: The descriptions for raw image upload need to be more elaborate. For example with the given explanations I am not certain how to upload to an azure container and it is not immediately obvious that the account given is tied to the cloud framework account that is associated with the users mash account. For example the account explanation could read as follows:

"Specify the cloud framework account as configured with the mash account add client command when the mash user was setup. The credentials associated with this cloud framework account will be used for the upload"

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
 
 azure_job_message = copy.deepcopy(base_job_message)
 azure_job_message['properties']['emails'] = string_with_example(
-    'test@fake.com'
+    'test@fake.com',
+    description='A comma-separated list of email addresses to be notified '
+                'of the progress of the publishing operation.'
```
rjschwei, line 28: s/publishing operation/job/

e-mail is not only sent when a job runs through publishing

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
+                'multiple disk generations.'
 )
-azure_job_message['properties']['offer_id'] = string_with_example('leap')
-azure_job_message['properties']['publisher_id'] = string_with_example('suse')
-azure_job_message['properties']['sku'] = string_with_example('15')
-azure_job_message['properties']['generation_id'] = string_with_example('gen2')
 azure_job_message['properties']['cloud_image_name_generation_suffix'] = string_with_example(
-    'gen2'
+    'gen2',
+    description='The suffix is appended to the cloud image name.'
```
rjschwei, line 55: s/is//

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
 azure_job_message['properties']['vm_images_key'] = string_with_example(
-    'microsoft-azure-corevm.vmImagesPublicAzure'
+    'microsoft-azure-corevm.vmImagesPublicAzure',
+    description='Dictionary key where sku information is found in the '
+                'offer doc.'
 )
-azure_job_message['properties']['publish_offer'] = {'type': 'boolean'}
+azure_job_message['properties']['publish_offer'] = {
+    'type': 'boolean',
+    'description': 'Whether to start the image publish process.'
```
rjschwei, line 64: s/publish/publishing/

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
+}
 azure_job_message['properties']['cloud_account'] = string_with_example(
-    'account1'
+    'account1',
+    description='The name of the cloud account credentials to use for image '
```
rjschwei, line 68: s/credentials//

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
+                'publishing.'
 )
 azure_job_message['properties']['source_container'] = string_with_example(
-    'container1'
+    'container1',
+    description='The ARM storage container where the image will be uploaded.'
```
rjschwei, line 73: Suggest:
"The name of the ARM (Azure Resource Manager) storage container where the image will be uploaded"

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
 )
 azure_job_message['properties']['source_resource_group'] = string_with_example(
-    'res_group1'
+    'res_group1',
+    description='The resource group where the source container and source '
+                'storage account reside.'
```
rjschwei, line 78: Suggest:
"The name of the resource group that contains the image storage container and storage account."

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
 )
 azure_job_message['properties']['source_storage_account'] = string_with_example(
-    'storage_account1'
+    'storage_account1',
+    description='The storage account where the source container resides.'
```
rjschwei, line 82: Suggest:
"The name of the storage account with which the image storage container is associated."

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
 )
 azure_job_message['properties']['destination_container'] = string_with_example(
-    'container2'
+    'container2',
+    description='The ASM (legacy) storage container where the image will '
+                'be copied to after image has passed testing.'
```
rjschwei, line 87: This needs a more detailed explanation to answer the question "Why is the image copied there?". For testing jobs this is not required for example.

mash/services/api/schema/jobs/azure.py
```diff
@@ -23,44 +23,83 @@
...
 )
 azure_job_message['properties']['destination_resource_group'] = string_with_example(
-    'res_group2'
+    'res_group2',
+    description='The resource group where the destination container and '
+                'destination storage account reside.'
```
rjschwei, line 92: Either here or in the "source" settings we need an explanation what "source" and "destination" mean

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
     'format': 'regex',
     'pattern': '^[0-9]{12}(,[0-9]{12})*$|^(all|none)$',
     'example': '123456789012,098765432109',
-    'examples': ['all', 'none', '123456789012,098765432109']
+    'examples': ['all', 'none', '123456789012,098765432109'],
+    'description': 'Sharing image to all shares the image publicly. '
```
rjschwei, line 30: s/all/"all"/

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+                   'Sharing to none keeps the image private and sharing '
```
rjschwei, line 31: s/none/"none"/

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
 ec2_job_account = {
     'type': 'object',
     'properties': {
-        'name': string_with_example('account1'),
-        'region': string_with_example('us-east-1'),
-        'root_swap_ami': string_with_example('ami-1234567890'),
-        'subnet': string_with_example('subnet-12345678')
+        'name': string_with_example(
+            'account1',
+            description='Name of cloud account.'
```
rjschwei, line 41: as associated with the mash user account

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+        ),
+        'region': string_with_example(
+            'us-east-1',
+            description='Region to use for initial image creation and '
+                        'instance testing.'
```
rjschwei, line 46: s/instance//

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+        ),
+        'root_swap_ami': string_with_example(
+            'ami-1234567890',
+            description='The image AMI to use for the root swap image upload '
+                        'method.'
+        ),
+        'subnet': string_with_example(
+            'subnet-12345678',
+            description='The subnet to use for image testing and image '
+                        'upload.'
```
rjschwei, line 56: s/upload/creation/

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+            description='Name of cloud account.'
+        ),
+        'region': string_with_example(
+            'us-east-1',
+            description='Region to use for initial image creation and '
+                        'instance testing.'
+        ),
+        'root_swap_ami': string_with_example(
+            'ami-1234567890',
+            description='The image AMI to use for the root swap image upload '
```
rjschwei, line 50: s/upload/creation/

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+                   'optional. If supplied region, root_swap_ami and '
+                   'subnet will override the account default values.'
 }
 
 ec2_job_message = copy.deepcopy(base_job_message)
 ec2_job_message['properties']['share_with'] = share_with
-ec2_job_message['properties']['allow_copy'] = {'type': 'boolean'}
+ec2_job_message['properties']['allow_copy'] = {
+    'type': 'boolean',
+    'description': 'Whether to allow copy of the image.'
```
rjschwei, line 71: s/image/snapshot with which the image is associated./

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
+}
 ec2_job_message['properties']['cloud_accounts'] = {
     'type': 'array',
     'items': ec2_job_account,
     'minItems': 1,
-    'example': [{'name': 'account1', 'region': 'us-east-2'}]
+    'example': [{'name': 'account1', 'region': 'us-east-2'}],
+    'description': 'A list of cloud account dictionaries. Either a '
+                   'cloud_account, cloud_accounts or cloud_groups is '
+                   'required for an EC2 job.'
```
rjschwei, line 90: Suggest to add:
"This allows the creation and testing of the same image in multiple partitions."

mash/services/api/schema/jobs/ec2.py
```diff
@@ -26,43 +26,83 @@
...
 }
 ec2_job_message['properties']['cloud_account'] = string_with_example(
-    'account1'
+    'account1',
+    description='The name of the cloud account credentials to use for '
```
rjschwei, line 94: s/credentials//

Suggest to add:
"This is mutually exclusive with the "cloud_acconts" setting."

mash/services/api/schema/jobs/gce.py
```diff
@@ -22,33 +22,59 @@
...
-    'testaccount1'
+    'testaccount1',
+    description='The account to use for launching and testing an instance '
+                'of the image. This is required if the cloud_account is a '
+                'publishing account which cannot launch instances.'
 )
 gce_job_message['properties']['cloud_account'] = string_with_example(
-    'account1'
+    'account1',
+    description='The name of the cloud account credentials to use for image '
```
rjschwei, line 64: s/credentials//

mash/services/api/schema/jobs/__init__.py
```diff
@@ -25,13 +25,26 @@
...
     'type': 'object',
     'properties': {
-        'package_name': string_with_example('kernel-default'),
-        'version': string_with_example('4.13.1'),
-        'release': string_with_example('1.1'),
+        'package_name': string_with_example(
+            'kernel-default',
+            description='The name of the package for this condition. '
+                        'If no name is provided for a condition then the '
+                        'condition is checked against the image itself.'
```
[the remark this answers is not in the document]
smarlowucf: This is a schema for a single condition. The description for the list of conditions is found here https://github.com/SUSE-Enceladus/mash/pull/646/files#diff-e0732d853d52c5e361b87f78f87c0d31R116 .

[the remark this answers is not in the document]
smarlowucf: I think it would be confusing to mention multiple conditions here

mash/services/api/schema/jobs/azure.py
smarlowucf: No these emails are specific to the Azure cloud partner portal. So this is indeed only updates related to publishing of Azure jobs.

Not to be confused with notification_email which is a base job attribute specific to mash.

mash/services/api/schema/jobs/__init__.py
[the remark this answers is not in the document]
rjschwei: Hi, I understand that this describes 1 condition, but while reading the thought of "how do I specify multiple conditions" is inevitable. We should not let that go unanswered and let the user keep wondering to eventually discover the answer elsewhere. Thus we can solve this making a reference to the  `conditions` section of the job doc or we can can do this by stating "multiple conditions can be specified". It is about not leaving the user hanging.

mash/services/api/schema/jobs/azure.py
rjschwei: Does this mean those e-mails come from Azure? If yes that should be stated to avoid the confusion I obviously fell into.

mash/services/api/schema/jobs/__init__.py
```diff
@@ -68,15 +90,33 @@
                 'publisher',
                 'deprecation'
             ],
-            'example': 'create'
+            'example': 'create',
+            'description': 'The last service in the pipeline to be executed. '
+                           'All services except the OBS service are valid '
+                           'values.'
```
rjschwei, line 96: Somehow we should still make the names of valid values available to the user. We should not make the user guess whether the service is named "test" or "testing" for example. I think listing those names here is logical but I can also see how something that gets configured on the server should not be in a hard coded list in the description. Maybe we can insert those names in the API call that produces the descriptions for the user?

mash/services/api/schema/jobs/__init__.py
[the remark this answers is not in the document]
smarlowucf: I think the naming is causing confusion here. So I will rename this to image_condition. A user pulling the definitions from the new mash client api will never see the description here. They will see the description in 'conditions' which is defined as a list of conditions. If a user pulls the raw json schema then they would see this description but it would be nested under 'conditions'.

Example output of annotated (default) response from mash client:

```
"conditions": {
        "description": "A list of image conditions to check the image against. Providing only the package name will ensure the package is in the image. The version is the package version from the build service whereas the release is the build or release number. At least one of package_name, version or release is required. If no package name is provided then the condition is against the image itself. Valid condition operators are >, <, >=, <=, or ==.",
        "example": [
            {
                "condition": ">=",
                "package_name": "kernel-default",
                "version": "4.13.1"
            }
        ],
        "type": "list of objects",
        "properties": {
            "condition": {
                "description": "The expression to use for comparing version and/or release.",
                "enum": [
                    ">=",
                    "==",
                    "<=",
                    ">",
                    "<"
                ],
                "example": "==",
                "type": "string"
            },
            "package_name": {
                "description": "The name of the package for this condition. If no name is provided for a condition then the condition is checked against the image itself.",
                "example": "kernel-default",
                "type": "string"
            },
            "release": {
                "description": "The build (release) number for the package or image. If no package_name is provided with the condition then the condition is against the image.",
                "example": "1.1",
                "type": "string"
            },
            "version": {
                "description": "The package or image version from the build service. If no package_name is provided with the condition then the condition is against the image.",
                "example": "4.13.1",
                "type": "string"
            }
        }
    },
```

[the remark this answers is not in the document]
smarlowucf: If we want to make changes then it should be done to the description for `conditions` not the description for `image_conditions`.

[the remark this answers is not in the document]
smarlowucf: It may be easier to see the entire code here.

```
image_condition = {
    'type': 'object',
    'properties': {
        'package_name': string_with_example(
            'kernel-default',
            description='The name of the package for this condition. '
                        'If no name is provided for a condition then the '
                        'condition is checked against the image itself.'
        ),
        'version': string_with_example(
            '4.13.1',
            description='The package or image version from the build service.'
                        ' If no package_name is provided with the condition '
                        'then the condition is against the image.'
        ),
        'release': string_with_example(
            '1.1',
            description='The build (release) number for the package or image.'
                        ' If no package_name is provided with the condition '
                        'then the condition is against the image.'
        ),
        'condition': {
            'type': 'string',
            'enum': ['>=', '==', '<=', '>', '<'],
            'example': '==',
            'description': 'The expression to use for comparing version '
                           'and/or release.'
        }
    },
    'additionalProperties': False,
    'anyOf': [
        {'required': ['package_name']},
        {'required': ['version']},
        {'required': ['release']}
    ]
}

...

        'conditions': {
            'type': 'array',
            'items': image_condition,
            'minItems': 1,
            'example': [
                {
                    'package_name': 'kernel-default',
                    'version': '4.13.1',
                    'condition': '>='
                }
            ],
            'description': 'A list of image conditions to check the image '
                           'against. Providing only the package name will '
                           'ensure the package is in the image. The version '
                           'is the package version from the build service '
                           'whereas the release is the build or release '
                           'number. At least one of package_name, version or '
                           'release is required. If no package name is '
                           'provided then the condition is against the image '
                           'itself. Valid condition operators are >, <, >=, '
                           '<=, or ==.'
        },
```

image_condition is just a variable. The user never sees this. Even if they pull raw schema they only see:

```
"conditions": {
            "description": "A list of image conditions to check the image against. Providing only the package name will ensure the package is in the image. The version is the package version from the build service whereas the release is the build or release number. At least one of package_name, version or release is required. If no package name is provided then the condition is against the image itself. Valid condition operators are >, <, >=, <=, or ==.",
            "example": [
                {
                    "condition": ">=",
                    "package_name": "kernel-default",
                    "version": "4.13.1"
                }
            ],
            "items": {
                "additionalProperties": false,
                "anyOf": [
                    {
                        "required": [
                            "package_name"
                        ]
                    },
                    {
                        "required": [
                            "version"
                        ]
                    },
                    {
                        "required": [
                            "release"
                        ]
                    }
                ],
                "properties": {
                    "condition": {
                        "description": "The expression to use for comparing version and/or release.",
                        "enum": [
                            ">=",
                            "==",
                            "<=",
                            ">",
                            "<"
                        ],
                        "example": "==",
                        "type": "string"
                    },
                    "package_name": {
                        "description": "The name of the package for this condition. If no name is provided for a condition then the condition is checked against the image itself.",
                        "example": "kernel-default",
                        "minLength": 1,
                        "type": "string"
                    },
                    "release": {
                        "description": "The build (release) number for the package or image. If no package_name is provided with the condition then the condition is against the image.",
                        "example": "1.1",
                        "minLength": 1,
                        "type": "string"
                    },
                    "version": {
                        "description": "The package or image version from the build service. If no package_name is provided with the condition then the condition is against the image.",
                        "example": "4.13.1",
                        "minLength": 1,
                        "type": "string"
                    }
                },
                "type": "object"
            },
            "minItems": 1,
            "type": "array"
        },
```

Hopefully that clarifies things?

mash/services/api/schema/jobs/__init__.py
smarlowucf: The enum is provided to the user with the annotated output. Isn't it duplicative to have the same list in description?

```
"last_service": {
        "description": "The last service in the pipeline to be executed. All services except the OBS service are valid values.",
        "enum": [
            "uploader",
            "create",
            "testing",
            "raw_image_uploader",
            "replication",
            "publisher",
            "deprecation"
        ],
        "example": "create",
        "type": "string",
        "required": true
    },
```

PR #648, opened 2020-04-20
Add uefi and secure boot options for testing.

smarlowucf:

Also add image project option required for GCE testing with the GCE API.

### Additional Information

When img-proof PR https://github.com/SUSE-Enceladus/img-proof/pull/242 gets merged and cut to a new version the mash version requirements should be updated.

### DIFF
--- a/examples/messages/gce_job.json
+++ b/examples/messages/gce_job.json
@@ -22,5 +22,6 @@
   "family": "sles-15",
   "tests": ["test_stuff"],
   "notification_email": "test@fake.com",
-  "guest_os_features": ["UEFI_COMPATIBLE"]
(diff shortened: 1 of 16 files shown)


rjschwei:
Hmm,

Missing the logic where we test the image twice, once with secure boot and once without. In a general sense being able to test just one or the other is important. For our use case of mash we need to be able to test both on the same image.

Did I just not see it?

smarlowucf:
There is no logic to test an image twice at this point. I think the first step is add the functionality so mash can be used to test secure boot. Then it would be good to have a discussion on how best we can handle the multiple image tests.

rjschwei:
Looping for uefi & non-uefi test will be added in subsequent change set.

PR #654, opened 2020-05-14
Add additional upload option and raw tarball upload for Azure

smarlowucf:

Update Azure utils to be more generic. For all blob methods accept both page and block blob usage. And for upload method accept block files or expanded image uploads.

Add additional uploads option. To provide a way for image checcksum and/or signature to be
uploaded alongside the image.

Add class for raw Azure image upload. This will upload an Azure image as a tarball without expansion. The class also allows for multi file upload to upload checksum and/or signature files.

Fix new flake8 warnings. Format string arg is no longer used.

Unit and E2E tests.

### DIFF
--- a/examples/messages/azure_job.json
+++ b/examples/messages/azure_job.json
@@ -34,5 +34,6 @@
   "vm_images_key": "key123",
   "notification_email": "test@fake.com",
   "publish_offer": true,
-  "cleanup_images": true
+  "cleanup_images": true,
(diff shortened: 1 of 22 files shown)


smarlowucf:
Re-based with fix for new flake8 warnings.

smarlowucf:
Re-based again to pull in loggerAdapter changes from master.

PR #655, opened 2020-05-15
Add snapshot sharing for EC2 images

smarlowucf:

Add option to skip replication in EC2. If skipping replication only provide the source region in target
region list.

Add utility method to share image snapshot. The method finds an image base on name and shares the related snapshot with a list of accounts.

Implement option to share image snapshot. In ec2 jobs the image snapshot can now be shared with a list of accounts. This allows the snapshot to be shared as part of a normal publishing job or as a snapshot share only job.

Unit and E2E tests.

### DIFF
--- a/examples/messages/job.json
+++ b/examples/messages/job.json
@@ -25,5 +25,6 @@
   "notification_email": "test@fake.com",
   "conditions_wait_time": 500,
   "disallow_licenses": ["MIT"],
-  "disallow_packages": ["*-mini"]
+  "disallow_packages": ["*-mini"],
+  "share_snapshot_with": "123456789012"
 }
--- a/examples/messages/job.json
+++ b/examples/messages/job.json
@@ -26,5 +26,5 @@
   "conditions_wait_time": 500,
   "disallow_licenses": ["MIT"],
   "disallow_packages": ["*-mini"],
-  "share_snapshot_with": "123456789012"
+  "share_snapshot_with": "123456789012,987654321210"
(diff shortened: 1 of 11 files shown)


rjschwei:
Confused, appears to have a partial implementation of replication skipping as well?

smarlowucf:
> Confused, appears to have a partial implementation of replication skipping as well?

It's not partial it's required to cover the marketplace only route. Marketplace only jobs are not replicated, only the snapshot in source region is shared with a list of accounts.

rjschwei:
> > Confused, appears to have a partial implementation of replication skipping as well?
> 
> It's not partial it's required to cover the marketplace only route. Marketplace only jobs are not replicated, only the snapshot in source region is shared with a list of accounts.

Bug job.json has no entry for "'skip_replication", did we leave out all the optional parameters?

smarlowucf:
> Bug job.json has no entry for "'skip_replication", did we leave out all the optional parameters? But "share_snapshot_with" should also be optional, thus should also not be in job.json.

Well there's two job.json's, mainly these are used by tests so args are added as needed by the tests to get the proper test coverage. I realize one of the json files is in a dir called examples but these examples are mostly stale. I just happen to sometimes update the one in examples when I update the one used by tests. Examples of job docs attrs are covered by the API service now so if it causes confusion I'd rather delete the examples directory.

smarlowucf:
Re-based with master to pick up flake8 fix.

PR #656, opened 2020-05-18
Fix logging in listener service classes.

smarlowucf:

Use logger adapter to pass in the context (job id and iteration) instead of a wrapper method. This allows job classes to use logger directly and pass the logger to any external tools used.


smarlowucf:
Re-based with master to pick up flake8 fix.

PR #658, opened 2020-05-20
Cleanup on EC2 create failure

smarlowucf:

If any accounts fail try to cleanup any images in the accounts that succeeded to make retrying job easier.

Unit tests.


smarlowucf:
Re-based again to pull in loggerAdapter changes from master.

PR #661, opened 2020-05-22
Add flag in ec2 testing.

smarlowucf:

If account is not in aws public and the image architecture is aarch64 skip testing as there are no instance types available.


smarlowucf:
> Fair, we just have to keep an eye on the announcements from AWS and remove the code when aarch64 becomes available there.

Ack, I can add a todo to ensure it doesn't get lost.

PR #662, opened 2020-05-29
Remove notification emails option.

smarlowucf:

For Azure publishing jobs. This is no longer used by the new partner center.

Unit tests.

### Additional Information

https://docs.microsoft.com/en-us/azure/marketplace/cloud-partner-portal-migration-faq#will-the-cloud-partner-portal-rest-apis-be-supported-post-migration

> For migrated offers, we'll no longer send notifications to the list of emails specified in the requests. Instead, the API service will align with the notification email process in Partner Center to send emails. Specifically, notifications will be sent to the email address set in the Seller contact info section of your Account settings in Partner Center, to notify you of operation progress.

### DIFF
--- a/examples/messages/azure_job.json
+++ b/examples/messages/azure_job.json
@@ -24,7 +24,6 @@
   "distro": "sles",
   "instance_type": "Basic_A2",
   "tests": ["test_stuff"],
-  "emails": "jdoe@fake.com",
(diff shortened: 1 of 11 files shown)


smarlowucf:
> Should be a major version bump for next release due to the API change.

Ack

PR #676, opened 2020-07-06
Use user email for notifications.

smarlowucf:

Instead of requiring the email as job arg. To enable notification emails notification_type now is required.

Unit tests.

### DIFF
--- a/examples/messages/azure_job.json
+++ b/examples/messages/azure_job.json
@@ -31,7 +31,7 @@
   "generation_id": "image-gen2",
   "cloud_image_name_generation_suffix": "gen2",
   "vm_images_key": "key123",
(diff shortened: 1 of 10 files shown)


bear454:
Right now, a user can have _someone else_ informed of job results. Is that feature not intentional?

rjschwei:
> Right now, a user can have _someone else_ informed of job results. Is that feature not intentional?

What was the answer to this question. I take it was discussed via chat?

I think having the ability to notify someone else is quite useful.

smarlowucf:
This feature was not intended however I have no opinion either way. I've added back the option but it is no longer a required attr. By default it will use `user.email`.

PR #678, opened 2020-07-20
Change message flow from push to pull.

smarlowucf:

The listener messages will be published to the service exchange sending the message so a single message can also be pulled/duplicated by the status service.

Unit and integration.

### DIFF
--- a/mash/services/listener_service.py
+++ b/mash/services/listener_service.py
@@ -93,7 +93,7 @@ class ListenerService(MashService):
             self.service_exchange, self.job_document_key, self.service_queue
         )
         self.bind_queue(
-            self.service_exchange, self.listener_msg_key, self.listener_queue
+            self.prev_service, self.listener_msg_key, self.listener_queue
         )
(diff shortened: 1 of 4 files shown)


smarlowucf:
Going to append subsequent commits here as I get them split out. As they all build on each other.

rjschwei:
> Going to append subsequent commits here as I get them split out. As they all build on each other.

Can we not have following PRs to the staging branch you created? Was just about to check this off

smarlowucf:
> > Going to append subsequent commits here as I get them split out. As they all build on each other.
> 
> Can we not have following PRs to the staging branch you created? Was just about to check this off

For commits that don't overlap it's easy. For commits that are tied together opening separate PR's can cause a lot of overhead handling merge/rebase as things are pushed to staging. Especially when changes are requested which affect the subsequent requests. :thinking:

rjschwei:
Fair, as long as it remains digestible

PR #680, opened 2020-07-23
Listener msgs

smarlowucf:

- Move knowledge of specific listener message args to job classes.
- Instead of setting attrs in listener_service this allows the args to be different between cloud frameworks.
- This is useful for providing status based on cloud.
- It also decouples the specifics of job classes from the service itself.
- Rename the attr from source_regions which made sense only for EC2 to status_msg. This dictionary will hold all listener msg info as a hash and can be different per cloud framework.
- Consume_queue method now requires exchange name as arg.

Unit and integration testing.

### DIFF
--- a/mash/services/create/azure_job.py
+++ b/mash/services/create/azure_job.py
@@ -49,8 +49,8 @@ class AzureCreateJob(MashJob):
         self.status = SUCCESS
         self.log_callback.info('Creating image.')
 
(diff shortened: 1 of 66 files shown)


smarlowucf:
Marked as work in progress. There's some changes I have queued that need to be organized with this prior to review/merge.

smarlowucf:
Okay this is ready for review.

PR #697, opened 2020-08-14
Add MASH cleanup service

jgleissner:
This PR adds a service that purges old image files.

This is intended for a first review only. I haven't implemented any tests yet nor the systemd unit file. Also, I didn't find a way to format the multi-line if statement in `mash/services/cleanup/service.py` in a way so that flake8 does not complain. Hints welcome.


mash/services/cleanup/config.py
```diff
@@ -0,0 +1,70 @@
...
+
+    * /etc/mash/mash_config.yaml
+
+    The mash configuration file is a yaml formatted file containing
+    information to control the behavior of the mash services.
+    """
+    def __init__(self, config_file=None):
+        super(CleanupConfig, self).__init__(config_file)
+
+    def get_download_directory(self):
```
smarlowucf, line 36: I think this could move to base config so it's not duplicated here and in obs service.

mash/services/cleanup/service.py
```diff
@@ -0,0 +1,72 @@
...
+        max_image_age = self.config.get_max_image_age()
+
+        if not os.path.isdir(download_dir):
+            self.log.error('Error: no such directory {}'.format(download_dir))
+            return
+
+        now = time.time()
+        cutoff = now - max_image_age * 86400
+
+        for entry in os.scandir(download_dir):
```
smarlowucf, line 67: Dir structure includes a dir per job id. Like:

```
/path/to/download_dir/
  job1/
    image.tar.gz
    image.tar.gz.sha256
  job2/
```

smarlowucf: We will want to delete the directories which will cleanup all artifacts for a given job.

smarlowucf: Also,could use the os.scandir function as a context manager so the resources are cleaned up when finished.

```
with os.scandir(download_dir) as scanner:
    for entry in scanner:
```

mash/services/cleanup/service.py
```diff
@@ -0,0 +1,72 @@
...
+            if (not entry.name.startswith('.')
```
smarlowucf, line 68: Docs make it sound like these are already ignored; `The entries are yielded in arbitrary order, and the special entries '.' and '..' are not included.`

mash/services/cleanup/config.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2020 SUSE Linux GmbH.  All rights reserved.
```
smarlowucf, line 1: SUSE LLC

mash/services/cleanup/defaults.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2020 SUSE Linux GmbH.  All rights reserved.
+#
```
rjschwei, line 2: SUSE LLC

mash/services/cleanup_service.py
```diff
@@ -0,0 +1,56 @@
...
+from mash.services.cleanup.service import CleanupService
+
+
+def main():
+    """
+    mash - cleanup service application entry point
+    """
+    try:
+        logging.basicConfig()
+        log = logging.getLogger('CleanupService')
```
rjschwei, line 35: This will create a giant log, i.e. every time the clean up service runs we append to the same log. which is general is fine, IMHO, but also means we should have some log rotation policy, based on size. We do not want to have to go fishing for an out of disk space condition in 3 or 4 years when we have long forgotten that the cleanup service writes to the same log.

smarlowucf: I believe the preference was to have system based log rotation as opposed to software level (in mash).

smarlowucf (approved):

rjschwei (approved):

PR #714, opened 2020-09-02
Update config and add wsgi template for db service

smarlowucf:
Update spec with new db service and move migrations to db service
from API service.


rjschwei (approved):
Other than the port question LGTM

jgleissner (approved):

PR #715, opened 2020-09-16
Move notification email to job creator.

smarlowucf:

The job creator is handling the job status and is the best spot to handle email notifs.

- Remove periodic email notification. This is not needed with status API. Only send an email when job finishes. No need to have the option of sending an email after each service finishes.
- Pop extra data from status message. Pop notification email and last service from status message before saving to DB. This is not helpful in the data section of job.

Unit and integration tests.

### DIFF
--- a/examples/messages/azure_job.json
+++ b/examples/messages/azure_job.json
@@ -31,7 +31,7 @@
   "generation_id": "image-gen2",
   "cloud_image_name_generation_suffix": "gen2",
   "vm_images_key": "key123",
-  "notification_type": "single",
+  "notify": true,
   "publish_offer": true,
   "cleanup_images": true,
   "additional_uploads": ["sha256"]
--- a/examples/messages/credentials_job.json
+++ b/examples/messages/credentials_job.json
@@ -7,6 +7,6 @@
         "last_service": "deprecation",
         "utctime": "now",
         "notification_email": "test@fake.com",
-        "notification_type": "single"
+        "notify": true
     }
 }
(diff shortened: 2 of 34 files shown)


jgleissner (approved):

examples/messages/azure_job.json
```diff
@@ -31,7 +31,7 @@
   "generation_id": "image-gen2",
   "cloud_image_name_generation_suffix": "gen2",
   "vm_images_key": "key123",
-  "notification_type": "single",
+  "notify": true,
```
rjschwei, line 34: This "breaks" what I remember we had discussed a long time ago in that a user could request to get notification as the job moves down the pipeline which with the implementation of status handling in the pipeline should now be possible, i.e.

```
"notification_type": "single"
```

means get one e-mail when the job is done

```
"notification_type": "all"
```

or something other than "all" means an e-mail per service completion. That, if I recall correctly was the intend of not making this a bool to begin with. I am not particularly attached to the idea but think it was a good plan to allow for notification at every step of the way and don't necessarily think taking this away (not that the "all" feature was implemented) is somewhat random.

smarlowucf: I don't think the "all" option was ever used. IMHO it was only a temporary solution until a status API was in place. I see the usefulness of maintaining email notif but not an email for every service. This is better handled by using the status API.

Also, because the email notif moves to job creator it is now in a stateless service. It doesn't have the knowledge it did when the email was sent from a given service in the pipeline. This means to properly handle send all and error notifications either the job creator becomes stateful or the services in pipeline add more state info to the status message.

Given these two things I chose to drop an unused and imho unnecessary feature to put preference on simpler code and maintaining no state where possible.

PR #719, opened 2020-10-16
Add use_build_time job flag.

smarlowucf:

If set the image build time will be used to format the cloud_image_name. If set and the build time is not available or the cloud_image_name does not have {date} format string an exception is raised.

Unit + E2E

Carefully


rjschwei (approved):

jgleissner (approved):

smarlowucf:
Rebased to fix conflict in requirements file.

PR #721, opened 2020-10-26
Add force replace image option for jobs.

smarlowucf:

If the flag is set to True and an existing tarball/image exists it will be deleted and replaced. Otherwise if the tarball/image already exists the job will fail.

Unit and E2E


smarlowucf:
Adding EC2 to this PR.

smarlowucf:
Ready for review.

jgleissner (approved):

rjschwei (approved):

PR #726, opened 2020-12-12
- Switch the default root volume for EC2 images from gp2

rjschwei:
(ssd) to   the new gp3 volume type. This affords users a 20% cost saving
  with no penalties. Requires ec2imgutils 9.0.1 or greater.

### DIFF
--- a/package/mash.spec
+++ b/package/mash.spec
@@ -40,7 +40,7 @@ BuildRequires:  python3-amqpstorm >= 2.4.0
 BuildRequires:  python3-APScheduler >= 3.3.1
 BuildRequires:  python3-python-dateutil >= 2.6.0
 BuildRequires:  python3-python-dateutil < 2.8.1
-BuildRequires:  python3-ec2imgutils >= 8.1.2
+BuildRequires:  python3-ec2imgutils >= 9.0.1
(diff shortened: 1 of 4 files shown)


jgleissner (approved):

package/mash.spec
```diff
@@ -40,7 +40,7 @@ BuildRequires:  python3-amqpstorm >= 2.4.0
 BuildRequires:  python3-APScheduler >= 3.3.1
 BuildRequires:  python3-python-dateutil >= 2.6.0
 BuildRequires:  python3-python-dateutil < 2.8.1
-BuildRequires:  python3-ec2imgutils >= 8.1.2
+BuildRequires:  python3-ec2imgutils >= 9.0.1
```
smarlowucf, line 43: There's also a requires clause.

smarlowucf (changes requested):
Also, want to update the .virtualenv.requirements.txt.

smarlowucf:
The check failure is from OCI. Will submit a separate PR to fix this.

smarlowucf:
Re-based to confirm build.

smarlowucf (approved):

PR #727, opened 2020-12-15
Fix cryptography install during testing.

smarlowucf:

Due to Oracle locking version requirements in the wrong place.


smarlowucf:
Merging to fix build.

PR #729, opened 2021-01-08
Update Azure SDK compute calls.

smarlowucf:

Use SDK version > 5 and latest azure identity package for client authentication.

Unit and integration testing.

Staged.

### Additional Information

E2E testing will be done once the staging branch is finished with all changes.

### DIFF
--- a/package/mash.spec
+++ b/package/mash.spec
@@ -28,7 +28,8 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
 BuildRequires:  python3-adal
-BuildRequires:  python3-azure-mgmt-compute < 5.0.0
+BuildRequires:  python3-azure-identity
+BuildRequires:  python3-azure-mgmt-compute >= 17.0.0
 BuildRequires:  python3-azure-mgmt-resource
 BuildRequires:  python3-azure-mgmt-storage
 BuildRequires:  python3-azure-storage-blob < 12.0.0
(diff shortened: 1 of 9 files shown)


package/mash.spec
```diff
@@ -57,7 +58,8 @@ BuildRequires:  python3-google-cloud-storage
 BuildRequires:  python3-google-api-python-client
 Requires:       rabbitmq-server
 Requires:       python3-adal
-Requires:       python3-azure-mgmt-compute < 5.0.0
+Requires:       python3-azure-identity
+Requires:       python3-azure-mgmt-compute
```
rjschwei, line 62: Shouldn't the condition now be ` >= 5.0.0` or did the newly used calls exist prior to 5.0.0?

rjschwei (approved):

package/mash.spec
smarlowucf: I need to dig further on that. I'm not sure when the azure-identity stuff was added. So it might be a newer version requirement than 5.0.0.

smarlowucf:
Sorry @jgleissner I mucked up the PR with the latest commit.

smarlowucf:
> Sorry @jgleissner I mucked up the PR with the latest commit.

Fixed.

jgleissner (approved):

package/mash.spec
smarlowucf: Best I can gauge is that the new `images.begin_create_or_update` method was added at 17.0.0 so I've updated the requirement accordingly.

PR #730, opened 2021-01-12
Migrate azure storage blob.

smarlowucf:

To latest Azure SDK >= 12.0.0. And remove replication service.

Unit and integration testing.

Staged

### Additional Information

E2E testing will be done once the staging branch is finished with all changes.


smarlowucf:
I dropped the:

`with create_json_file(credential) as auth_file:`

clause from a few spots which mucked up the diff. The only diff in those spots is wherever auth_file was used the credential/credentials dict is used instead.

rjschwei (approved):

PR #732, opened 2021-01-14
Handle firmware list for Azure.

smarlowucf:

Handle creation and testing of images in both uefi and bios (legacy) boot firmware. An image is created for each firmware in the boot_firmware list.

Unit tests.

Staged

### Additional Information

E2E once this final PR is merged to staging.

### DIFF
--- a/test/unit/services/create/azure_job_test.py
+++ b/test/unit/services/create/azure_job_test.py
@@ -1,6 +1,6 @@
 from pytest import raises
 from unittest.mock import (
-    MagicMock, Mock, patch
+    MagicMock, Mock, patch, call
 )
 
 from mash.services.create.azure_job import AzureCreateJob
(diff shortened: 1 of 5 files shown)


test/unit/services/create/azure_job_test.py
```diff
@@ -41,7 +41,8 @@ def setup(self):
             'container': 'container',
             'storage_account': 'storage',
             'region': 'region',
-            'cloud_image_name': 'name'
+            'cloud_image_name': 'name',
+            'boot_firmware': ['bios', 'uefi']
```
rjschwei, line 45: @KeithMnemonic the mashouse data set needs to account for this new seeting for SLES 15 GA and later images.

rjschwei (approved):

jgleissner (approved):

PR #735, opened 2021-02-23
Add helper error message when tests fail.

smarlowucf:

Add message to errors list which will be visible in notification email and job status.


smarlowucf:
@rjschwei good?

rjschwei (approved):

PR #747, opened 2021-03-15
Add Aliyun account model.

smarlowucf:

Add database models, routes and utils for Aliyun accounts.

Unit tests.


smarlowucf:
Sorry this is a lot of code. There's quite a bit of copy and pasting that can be done given functions are all the same with different names and args.

rjschwei (approved):

jgleissner (approved):

PR #764, opened 2021-03-31
Update default user for Aliyun.

smarlowucf:
SSH user name is ali-user.


smarlowucf:
Added a fix for the gce test. The latest version of google api client broken the httperror usage.

rjschwei (approved):

PR #771, opened 2021-05-07
Fix sas url for azure publish.

smarlowucf:

SAS URL should have container permissions for read and write.


rjschwei (approved):

bear454:
I'd like to see a more comprehensive test then using a completely fake SAS URL.

smarlowucf:
> I'd like to see a more comprehensive test then using a completely fake SAS URL.

The mock is only mocking the sas token generation which is a string of random characters. But I understand the tests are not great :(.

PR #773, opened 2021-05-20
Fixed spec file requirement

schaefi:
Mash sets a max version in the spec file for the dateutil
module. However, the implementation doesn't require this
constraint as it works with any modern version of dateutil.
In an effort to make mash being able to build and work in
a variety of linux versions such max version constraints
will be roadblocks in moving forward and should be deleted
if possible.


smarlowucf:
As far as I can tell there was no reason to have the version limit.

smarlowucf (approved):

PR #774, opened 2021-05-20
Always pass in jwt algorithms for decoding

schaefi:
Starting with versions >= 2.0.0 of pyjwt the setting of the
decoding algorithms became mandatory.


smarlowucf (approved):
This looks good for now. Mash does have a jwt algorithm config option but it's not used in the code right now so the default HS256 is being used when creating tokens.

PR #780, opened 2021-06-21
Add a slow rollout policy to new GCE images.

illfelder:
Introduce slow rollout policy for new GCE image publication and deprecation.

TBD (Please help me figure out how to test these changes work in your build environment).

These changes should be used whenever publishing new images in GCE. No action should be required to use these new changes.

### Additional Information
Existing usage of the API should behave as normal and without change.

### DIFF
--- a/mash/services/deprecate/gce_job.py
+++ b/mash/services/deprecate/gce_job.py
@@ -63,7 +63,7 @@ class GCEDeprecateJob(MashJob):
 
         try:
             project = credential.get('project_id')
-            compute_driver = get_gce_compute_driver(credential)
+            compute_driver = get_gce_compute_driver(credential, version='alpha')
 
             deprecate_gce_image(
                 compute_driver,
--- a/mash/utils/gce.py
+++ b/mash/utils/gce.py
@@ -17,6 +17,7 @@
 #
 
 import datetime
+import itertools
 import random
 import time
 
@@ -75,6 +76,42 @@ def get_region_list(compute_driver, project):
     return region_names
 
 
+def get_zones(compute_driver, project):
+    """
+    Returns a list of zones for a project.
+    """
+    zones = compute_driver.zones().list(project=project).execute()
+    zones = sorted(zones.get('items'), key=lambda zone: zone.get('name'))
+    zones_map = {}
+    for zone in zones:
+        region = zone.get('region')
+        name = zone.get('name')
+        zones_map.setdefault(region, []).append(name)
+    zones = list(zones_map.values())
+    random.shuffle(zones)
+    return [
+        'zones/%s' % zone for zone in itertools.chain(
+            *itertools.zip_longest(*zones)) if zone is not None]
+
+
+def create_gce_rollout(compute_driver, project):
+    """
+    Create a rollout policy for publishing and deprecating images.
+    """
+    format_str = '%Y-%m-%dT%H:%M:%SZ'
+    now = datetime.datetime.now()
+    zones = zones = get_zones(compute_driver, project)
+    policies = {}
+    for num, zone in enumerate(zones):
+        rollout_time = now + datetime.timedelta(hours=num)
+        policies[zone] = rollout_time.strftime(format_str)
+    default = now + datetime.timedelta(hours=len(zones))
+    return {
+        'defaultRolloutTime': default.strftime(format_str),
+        'locationRolloutPolicies': policies
+    }
+
+
 def create_gce_image(
     compute_driver,
     project,
@@ -82,7 +119,8 @@ def create_gce_image(
     cloud_image_description,
     blob_uri,
     family=None,
-    guest_os_features=None
+    guest_os_features=None,
+    rollout=None
 ):
     """
     Create a GCE framework image for the blob.
@@ -92,12 +130,11 @@ def create_gce_image(
     """
     kwargs = {
         'name': cloud_image_name,
+        'family': family,
         'description': cloud_image_description,
-        'rawDisk': {'source': blob_uri}
+        'rawDisk': {'source': blob_uri},
+        'rolloutOverride': rollout
     }
-
-    if family:
-        kwargs['family'] = family
 
     if guest_os_features:
         kwargs['guestOsFeatures'] = [
@@ -198,11 +235,15 @@ def deprecate_gce_image(
         replacement_image_name
     )
     replacement_image_uri = replacement_image['selfLink']
+    # Image deprecation should follow the same rollout policy as the image it is
+    # replacing.
+    rollout = replacement_image.get('rolloutOverride')
 
     kwargs = {
         'replacement': replacement_image_uri,
         'deleted': delete_timestamp,
-        'state': 'DEPRECATED'
+        'state': 'DEPRECATED',
+        'stateOverride': rollout
     }
 
     compute_driver.images().deprecate(
@@ -230,7 +271,7 @@ def wait_on_image_ready(compute_driver, project, cloud_image_name):
         time.sleep(5)
 
 
-def get_gce_compute_driver(credentials):
+def get_gce_compute_driver(credentials, version='v1'):
     """
     Get an SDK compute driver based on credentials dictionary.
 
@@ -242,7 +283,7 @@ def get_gce_compute_driver(credentials):
 
     return discovery.build(
         'compute',
-        'v1',
+        version,
         credentials=client_creds,
         cache_discovery=False
     )
--- a/mash/utils/gce.py
+++ b/mash/utils/gce.py
@@ -87,8 +87,13 @@ def get_zones(compute_driver, project):
         region = zone.get('region')
         name = zone.get('name')
         zones_map.setdefault(region, []).append(name)
+    # Create a list of lists. Each sublist is the zone names in a different
+    # region.
     zones = list(zones_map.values())
+    # Permute the sublists so the region ordering is random, but the zone
+    # ordering is deterministic and sorted.
     random.shuffle(zones)
+    # Interleave varying sized lists of zones adding a 'zones/' prefix.
     return [
         'zones/%s' % zone for zone in itertools.chain(
             *itertools.zip_longest(*zones)) if zone is not None]
@@ -100,7 +105,7 @@ def create_gce_rollout(compute_driver, project):
     """
     format_str = '%Y-%m-%dT%H:%M:%SZ'
     now = datetime.datetime.now()
-    zones = zones = get_zones(compute_driver, project)
+    zones = get_zones(compute_driver, project)
     policies = {}
     for num, zone in enumerate(zones):
         rollout_time = now + datetime.timedelta(hours=num)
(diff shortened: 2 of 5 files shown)


illfelder:
@rjschwei - can you please loop in someone to take a look?

smarlowucf:
Hi @illfelder I will take a look but it will be sometime next week at the earliest.

mash/utils/gce.py
```diff
@@ -92,13 +130,12 @@ def create_gce_image(
...
         'name': cloud_image_name,
+        'family': family,
         'description': cloud_image_description,
-        'rawDisk': {'source': blob_uri}
+        'rawDisk': {'source': blob_uri},
+        'rolloutOverride': rollout
     }
 
-    if family:
-        kwargs['family'] = family
```
smarlowucf, line 100: We don't always include family. For example when testing.

mash/services/deprecate/gce_job.py
```diff
@@ -63,7 +63,7 @@ def run_job(self):
 
         try:
             project = credential.get('project_id')
-            compute_driver = get_gce_compute_driver(credential)
+            compute_driver = get_gce_compute_driver(credential, version='alpha')
```
smarlowucf, line 66: Does alpha rely on a specific SDK version? We are using distribution packages instead of PyPI so we aren't pulling in latest versions as they release.

mash/utils/gce.py
```diff
@@ -75,14 +76,51 @@ def get_region_list(compute_driver, project):
     return region_names
 
 
+def get_zones(compute_driver, project):
+    """
+    Returns a list of zones for a project.
+    """
+    zones = compute_driver.zones().list(project=project).execute()
+    zones = sorted(zones.get('items'), key=lambda zone: zone.get('name'))
```
smarlowucf, line 84: Why are zones sorted only to be shuffled later?

mash/utils/gce.py
```diff
@@ -75,14 +76,51 @@ def get_region_list(compute_driver, project):
...
+    zones_map = {}
+    for zone in zones:
+        region = zone.get('region')
+        name = zone.get('name')
+        zones_map.setdefault(region, []).append(name)
```
smarlowucf, line 89: We are going through a list of zones then re-packing them into a list for their respective regions. Can't we just go through a list of regions and pull a list of zones? I realize the performance difference probably isn't that much but I think it would make the function more readable.

mash/utils/gce.py
```diff
@@ -75,14 +76,51 @@ def get_region_list(compute_driver, project):
...
+    zones = list(zones_map.values())
+    random.shuffle(zones)
+    return [
```
smarlowucf, line 92: Is there a way to do this that's more readable?

mash/utils/gce.py
```diff
@@ -75,14 +76,51 @@ def get_region_list(compute_driver, project):
...
+            *itertools.zip_longest(*zones)) if zone is not None]
+
+
+def create_gce_rollout(compute_driver, project):
+    """
+    Create a rollout policy for publishing and deprecating images.
+    """
+    format_str = '%Y-%m-%dT%H:%M:%SZ'
+    now = datetime.datetime.now()
+    zones = zones = get_zones(compute_driver, project)
```
smarlowucf, line 103: Typo s/zones = zones =/zones =

smarlowucf (changes requested):

smarlowucf:
Please sign all the commits. See the GitHub docs for more info on signing commits and setting up a gpg key if necessary. https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits

mash/utils/gce.py
illfelder: Fields that are not set will be None which is ignored by the API.

mash/services/deprecate/gce_job.py
illfelder: It shouldn't, but if you run into issues let me know.

mash/utils/gce.py
illfelder: The zones are sorted, the regions are shuffled. This logic will always roll out `us-central1-a` before `us-central1-f`, but `us-central1` and `us-west1` have no predetermined ordering.

mash/utils/gce.py
illfelder: We wanted the zones ordering to be deterministic so that we know `-a` will roll out before `-b`.

mash/utils/gce.py
illfelder: Thanks for catching this!

mash/utils/gce.py
illfelder: It can be done in other ways, but I wouldn't consider them more readable. I added comments explaining the logic though.

mash/utils/gce.py
smarlowucf: If you get a list of regions, are the zones preloaded or lazy loaded when you call the zones attr on a region? I ask because it seems there's a significant difference in requests in each route.

The current implementation would have 1 request to get all zones then a request for every zone to get it's respective region. Whereas getting a list of regions, if the zones are preloaded is a single request.

mash/utils/gce.py
smarlowucf: Thanks for the comments. The itertools part is still quite confusing to me and I think that would be the same for anyone else who comes along the code in the future. Without digging up documentation for itertools I don't understand what the return type is. What does the output look like that's returned?

smarlowucf:
E2E testing is difficult. I think only we can do that unless you want to manually create a pipeline for testing. The unit tests and formatting checks would be next. All code has tests `pytest --cov=mash` and has no formatting errors `flake8 mash tests`.

mash/utils/gce.py
smarlowucf: I pulled the code out to follow logic so I understand the output now. This is confusing because it starts as a list of lists and the comment:

`Interleave varying sized lists of zones adding a 'zones/' prefix.`

Is a bit vague. To me it seemed like the end result was still a list of lists. Part of the confusion is because the format string contains `zones/%s` has a connotation of multiple not singular. I understand the difficulty of writing this in another way but if the different steps were split out into separate lines with more clear explanation on what's happening I think it will be more readable.

mash/utils/gce.py
```diff
@@ -75,14 +76,56 @@ def get_region_list(compute_driver, project):
...
+        zones_map.setdefault(region, []).append(name)
+    # Create a list of lists. Each sublist is the zone names in a different
+    # region.
+    zones = list(zones_map.values())
+    # Permute the sublists so the region ordering is random, but the zone
+    # ordering is deterministic and sorted.
+    random.shuffle(zones)
+    # Interleave varying sized lists of zones adding a 'zones/' prefix.
+    return [
+        'zones/%s' % zone for zone in itertools.chain(
```
smarlowucf, line 98: I missed this the first time through. Can we switch to the `.format` style here? I realize it's a code style thing but I've tried to remove any of the old style `%s` formatting from the codebase.

smarlowucf: Unfortunately we cannot do f' strings yet to support older python versions.

illfelder: Updated.

mash/utils/gce.py
illfelder: Can you take a look at the most recent version where I give examples of what each step should look like?

smarlowucf:
Looks good. Next items would be testing,formatting then E2E tests.

- pytest --cov=mash (passing plus full coverage)
- flake8 mash tests (no errors)
- E2E testing with production setup

illfelder:
- The pytest coverage hasn't changed (still 100% for everything).
- flake8 identified an import issue that I addressed and is now clean.

illfelder:
Issue I'm encountering while trying to run pytest. I verified all dev requirements are installed.
```
pytest --cov=mash
============================================================================ test session starts ============================================================================
platform linux -- Python 3.9.2, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /usr/local/google/home/illfelder/cloud/mash, configfile: setup.cfg
plugins: cov-2.12.1
collected 0 items / 1 error                                                                                                                                                 
Coverage.py warning: Module mash was never imported. (module-not-imported)
Coverage.py warning: No data was collected. (no-data-collected)
WARNING: Failed to generate report: No data to report.

/usr/local/google/home/illfelder/.local/lib/python3.9/site-packages/pytest_cov/plugin.py:271: PytestWarning: Failed to generate report: No data to report.

  self.cov_controller.finish()

================================================================================== ERRORS ===================================================================================
_______________________________________________________________________ ERROR collecting test session _______________________________________________________________________
/usr/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
../../../../.local/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
    exec(co, module.__dict__)
services/api/conftest.py:3: in <module>
    from mash.services.api.app import create_app
E   ModuleNotFoundError: No module named 'mash'

----------- coverage: platform linux, python 3.9.2-final-0 -----------

========================================================================== short test summary info ==========================================================================
ERROR ../.. - ModuleNotFoundError: No module named 'mash'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================================================= 1 error in 0.14s ==============================================================================
```

smarlowucf:
To get pytest to work properly you should be able to install mash as editable with the following:

`pip install -e .`

And to get the dev requirements:

`pip install -r .virtualenv.dev-requirements.txt`

illfelder:
Thank you. that did the trick! I've updated tests and they now fully pass.

smarlowucf:
Great! Looks good. The last piece is to get coverage on the new `get_zones` and `create_gce_rollout` functions. Note that the tests require manually mocking the GCE API calls so they run consistently without network connection.

`pytest --cov=mash --cov-report=term-missing`

smarlowucf (approved):

illfelder:
Additional test coverage added. Thanks for the command!

smarlowucf:
I tested this successfully with api client == 1.8.0 and in production we have 1.7.4. As far as I can tell there's no breaking changes between 1.8.0 and 1.7.4 so this should be good to merge.

smarlowucf:
Thanks for the contribution @illfelder :+1: , much appreciated. This will go out with the next release.

PR #782, opened 2021-07-02
Drop limit on idna

smarlowucf:

Old requirement to prevent incompatibility with requests.

### Additional Information

Reference: https://github.com/SUSE-Enceladus/mash/commit/fd612348bbb756a7a892ad88c139fae0668712f2


schaefi (approved):
This little change has such a great positive effect :+1:

PR #783, opened 2021-07-02
Drop version limit on werkzeug

smarlowucf:

This was added due to a temporary version conflict with Flask-restplus.

### Additional Information

Flask-restplus was replaced by Flask-restx.

Reference: https://github.com/SUSE-Enceladus/mash/commit/2ad81a85816418c7f1edb51c35ad069ebe415c12


schaefi (approved):
Looks great :+1: 

there is a conflict though

PR #787, opened 2021-07-27
Fix incompatibility in Flask > 2.0.0

smarlowucf:

The new async functionality in flask causes issues with the way
current_app was being mocked. Mock each object in the hierarchy
to get around asyn coroutines not being run.

Unit tests


smarlowucf:
@rjschwei good?

rjschwei (approved):

PR #791, opened 2021-12-03
- Support new region in EC2

rjschwei:

Support nes EC2 region ap-southeast-3


smarlowucf:
Thanks

PR #794, opened 2021-12-16
Add rpm-macros to build requirements in spec

smarlowucf:

### DIFF
--- a/package/mash.spec
+++ b/package/mash.spec
@@ -25,6 +25,7 @@ License:        GPL-3.0+
 Group:          System/Management
 Source:         mash-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+BuildRequires:  python-rpm-macros
 BuildRequires:  python3-devel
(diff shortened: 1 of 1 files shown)


package/mash.spec
```diff
@@ -27,6 +27,7 @@ Source:         mash-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
+BuildRequires:  python3-rpm-macros
```
rjschwei, line 30: Please use `python-rpm-macros`. `python3-rpm-macros` is just a `Provides` and as such if someone searches for `python3-rpm-macros` in the Build Service nothing will show up which will lead to an unnecessary goose chase.

rjschwei (changes requested):

package/mash.spec
smarlowucf: Rebased

rjschwei (approved):

PR #799, opened 2022-02-08
Add a 5 minute wait before checking on images

smarlowucf:
Replication can take time to finish in all the regions. This
prevents unnecessary API calls while the images process.


smarlowucf:
Relies on #800 to fix tests

rjschwei (approved):

PR #809, opened 2022-03-24
Implement EC2 MP publish class

smarlowucf:
With a start change set utility function. This function handles
submitting a new image version to the AWS Marketplace.

### DIFF
--- a/mash/services/jobcreator/ec2_job.py
+++ b/mash/services/jobcreator/ec2_job.py
@@ -50,9 +50,6 @@ class EC2Job(BaseJob):
             False
         )
 
-        if self.publish_in_marketplace:
-            self.last_service = 'publish'  # No deprecation for MP images
-
     def _get_target_regions_list(self):
         """
         Return list of region info.
@@ -106,7 +103,9 @@ class EC2Job(BaseJob):
                     'os_version': self.os_version,
                     'usage_instructions': self.usage_instructions,
                     'recommended_instance_type': self.recommended_instance_type,
-                    'publish_regions': self.get_publish_regions()
+                    'allow_copy': self.allow_copy,
+                    'share_with': self.share_with,
+                    'publish_regions': self.get_mp_publish_regions()
                 }
             }
         else:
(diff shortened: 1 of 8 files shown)


mash/services/jobcreator/ec2_job.py
```diff
@@ -106,7 +103,9 @@ def get_publish_message(self):
                     'os_version': self.os_version,
                     'usage_instructions': self.usage_instructions,
                     'recommended_instance_type': self.recommended_instance_type,
-                    'publish_regions': self.get_publish_regions()
+                    'allow_copy': self.allow_copy,
+                    'share_with': self.share_with,
+                    'publish_regions': self.get_mp_publish_regions()
```
rjschwei, line 108: This is always `us-east-1`, why not just hard code this condition to make it more obvious that MP publishing is "special"?

Or you could move the `if` condition into `get_publish_regions()` and make the `if` block much shorter such as

```
if self.publish_in_marketplace:
    return ['us-east-1']
```
at the beginning of `get_publish_regions()`.

Cannot follow the reasoning for your choices, can you elaborate?

Thanks

smarlowucf: An EC2 job can be submitted with multiple accounts. The accounts are referenced (in a dict) by a source region (as key) hence the name "publish_regions". It's not a list of regions it's a list of accounts indexed by source regions.

smarlowucf: Well actually for MP class the dictionary is inverted. But the naming convention is the same throughout the code (inverted or not). So changing it one place I think only makes things more confusing. If we want to change the name it should be changed everywhere which is a separate issue.

rjschwei (approved):

mash/services/jobcreator/ec2_job.py
smarlowucf: As a note, jobcreator takes the job info from the API and translates the data based on cloud and service to send info the format that is best ingested. This is different for each combination of cloud/service and thus the variable naming is confusing. As jobcreator was written before the API service existed and the API service shares similar functionality it's possible this part of job creator could be removed/consildated into the API. And the service classes could accept a standard message format.

PR #818, opened 2022-04-18
Revert publish date change from ec2 mp

smarlowucf:

Instead of passing publish date as a message argument which can be None, parse the date
timestamp from the image name for use in version title.

### DIFF
--- a/mash/services/create/ec2_job.py
+++ b/mash/services/create/ec2_job.py
@@ -90,7 +90,6 @@ class EC2CreateJob(MashJob):
             self.base_cloud_image_name,
             timestamp=timestamp
         )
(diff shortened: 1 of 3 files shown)


rjschwei:
Probably good to add some postfix to `'image_name_v20220202'` in the test to model the test our image names.  But this is somewhat cosmetic.

rjschwei (approved):

smarlowucf:
> Probably good to add some postfix to `'image_name_v20220202'` in the test to model the test our image names. But this is somewhat cosmetic.

Mash only requires the date string not the postfix. I think it's good as it is for now.

jesusbv (approved):
LGTM

PR #820, opened 2022-05-11
Add azure sig create class

smarlowucf:

And new azure job arguments to enable sig testing.

- gallery_name: references the shared image gallery in use
- gallery_resource_group: optional resource group where the shared
image gallery is found. By default source_resource_group is used.

### DIFF
--- a/mash/services/create/azure_sig_job.py
+++ b/mash/services/create/azure_sig_job.py
@@ -1,0 +1,124 @@
+# Copyright (c) 2022 SUSE LLC.  All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# mash is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with mash.  If not, see <http://www.gnu.org/licenses/>
+#
+
+import re
+
+from datetime import datetime
+
+from azure_img_utils.azure_image import AzureImage
+
+# project
+from mash.services.mash_job import MashJob
+from mash.mash_exceptions import MashCreateException
+from mash.services.status_levels import SUCCESS
+
+
+class AzureSIGCreateJob(MashJob):
+    """
+    Implements Azure shared image gallery image creation
+
+    Creates a new version of an image definition.
+    """
+    def post_init(self):
+        try:
+            self.container = self.job_config['container']
+            self.storage_account = self.job_config['storage_account']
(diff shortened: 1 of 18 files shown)


smarlowucf:
All tests passed.

mash/services/create/azure_sig_job.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def run_job(self):
+        self.status = SUCCESS
+        self.status_msg['images'] = {}
+        self.log_callback.info('Creating image.')
+
+        self.cloud_image_name = self.status_msg['cloud_image_name']
+        self.blob_name = self.status_msg['blob_name']
+
+        timestamp = re.findall(r'\d{8}', self.cloud_image_name)[0]
```
rjschwei, line 68: This is very SUSE specific, also exists in aliyun_job, gce_job, and ec2-mp_job. We should have an issue to find a more general solution that can also work with the SUSE naming convention.

And yes, I should have noticed that in previous code reviews, sorry.

mash/services/create/azure_sig_job.py
```diff
@@ -0,0 +1,124 @@
...
+from mash.services.mash_job import MashJob
+from mash.mash_exceptions import MashCreateException
+from mash.services.status_levels import SUCCESS
+
+
+class AzureSIGCreateJob(MashJob):
+    """
+    Implements Azure shared image gallery image creation
+
+    Creates a new version of an image definition.
```
rjschwei, line 35: The comment is somewhat misleading. In the documentation and the workflow the "image definition" is a containment that holds images. mash does not create this containment but expects the containment to exist. mash creates a new image version inside the "image definition" containment.

mash/services/test/azure_sig_job.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+
+        self.request_credentials([self.account], cloud='azure')
+        credentials = self.credentials[self.account]
+        self.image_version = self.status_msg['image_version']
+
+        with create_json_file(credentials) as auth_file:
+            for image_definition in self.status_msg['images']:
+                # TODO: Need a way to determine firmware from image_definition
+                boot_firmware = 'uefi' if 'gen2' in image_definition else 'bios'
```
rjschwei, line 113: boot firmware is also uefi for aarch64, but I am not certain if that gets ignored by the Azure API if the image is inside an aarch64 image definition.

rjschwei (approved):

mash/services/create/azure_sig_job.py
smarlowucf: Agreed, could be a config option. Also, the code (around date string handling) has become spread out across multiple services and has other SUSE idiosyncrasies that could be cleaned up.

smarlowucf: https://github.com/SUSE-Enceladus/mash/issues/821

mash/services/test/azure_sig_job.py
smarlowucf: Okay, this is solved for now. Realized that in img-proof we are already ignoring uefi and secure boot options for Azure. Because this value is set at image creation not instance creation.

PR #830, opened 2022-06-16
Prepare for aarch64 in GCE

rjschwei:

Image creation for GCE images adds a new architecture setting to distinguish
  between x86_64 and aarch64. Set the image attribute on image creation.

### Additional Information

We do not know the instance types yet as such invocation of img-proof will need another PR to handle testing.

### DIFF
--- a/test/unit/services/create/gce_job_test.py
+++ b/test/unit/services/create/gce_job_test.py
@@ -131,7 +131,9 @@ class TestGCECreateJob(object):
         self.complete_job_doc['cloud_architecture'] = 'aarch64'
         self.complete_job_doc['cloud_image_name'] = 'sles-15-sp4-v20210731'
         self.complete_job_doc['family'] = 'sles-15-arm64'
-        self.job = GCECreateJob(self.complete_job_doc, self.config)
+        job = GCECreateJob(self.complete_job_doc, self.config)
+        job._log_callback = Mock()
+        job.credentials = self.credentials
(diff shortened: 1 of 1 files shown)


mash/utils/gce.py
```diff
@@ -151,6 +152,7 @@ def create_gce_image(
         'description': cloud_image_description,
         'rawDisk': {'source': blob_uri},
         'rolloutOverride': rollout
```
smarlowucf, line 154: Syntax error

mash/utils/gce.py
```diff
@@ -151,6 +152,7 @@ def create_gce_image(
...
+        'architecture': arch.upper()
```
smarlowucf, line 155: This works in terms of Mash because we are injecting a driver that is set to alpha version. Given this code will be pulled into a utility I think it's okay to leave it. However, we may want keep in mind at least some error handling that directs a user what to do. E.g. if they inject a driver that is for version v1 and try to set architecture which doesn't not exist yet.

smarlowucf (changes requested):
Linting fails due to syntax error

smarlowucf:
If we switch the test to use `aarch64` arch then it will hit the new loc and fix the coverage back to 100%.

rjschwei:
Sorry for the churn, cannot run the tests locally and am not in the moot to dig into it ATM. I'll get it sorted out eventually.

PR #837, opened 2022-09-15
New AWS region

rjschwei:
- Update the configuration to include the new me-central-1 region

### DIFF
--- a/config/mash_config.yaml
+++ b/config/mash_config.yaml
@@ -46,6 +46,7 @@ cloud:
         - eu-west-1
         - eu-west-2
         - eu-west-3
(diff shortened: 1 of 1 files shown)


apozsuse:
LGTM :+1: 

I compared this file with the applied config and here's missing the 'ap-northeast-3' region. Should we include it here also?

apozsuse (approved):

smarlowucf:
Note this is only the example config. Also, the failure is due to changes in werkzeug. Yet again the flask-restx package is stuck in a not maintained state and does not have a new release to handle latest werkzeug.

PR #839, opened 2022-10-03
 Adds files to mock mash API

apozsuse:

This PR provides the required files to mock mash API in an easy way.

A test script is provided that makes requests to all mocked endpoints.

This code is not meant for production.


smarlowucf (commented):
I may have missed it but it would be nice to have instructions on how to keep the templates up-to-date. Otherwise we will end up with configuration drift between the mocked api and the actual API if an endpoint is changed, added or deleted.

Ideally this would become part of the development documentation as another required step when making API changes.

mock/Readme.md
```diff
@@ -0,0 +1,126 @@
...
+
+```sh
+sudo podman run -it \
+   -p 8000-8005:8000-8005 \
+   -v `pwd`:/tmp/ \
+   testrio/mockintosh \
+   /tmp/mash_mockintosh_config.yaml
+```
+
+By default, the fake API is listening in the port 8000.
```
rjschwei, line 47: s/in the/on/

mock/Readme.md
```diff
@@ -0,0 +1,126 @@
...
+├── auth
+├── jobs
+│   ├── aliyun
+│   ├── azure
+│   ├── ec2
+│   ├── gce
+└── user
+```
+
+2- Then go modify the template to match the modified behavior in the endpoint.
```
rjschwei, line 124: s/in/of/

rjschwei (changes requested):

rjschwei:
Generally I am wondering if I am missing something. Based on this I am thinking we are trying to test the response of the api front end to mash, i.e. what amounts to be the API to job-creator service. Why do we need to run a container that provides pre-canned responses? This would be something for testing the client, or not?

Meaning if I want to test https://github.com/SUSE-Enceladus/mash-client I can test it by implementing a mock API that runs in a a container and I'd point that container there.

But how does this test benefit mash (server side) development?
- it would find if an incompatibility gets introduced in the API, fair enough but is that worth the effort?

As documented when the API gets changed the test implementation also has to be touched and as such this doesn't really prevent the API change from being released without the client being touched. The case where we prevent the client from being broken would only be caught if we enforce client testing to be run against the mocks that are checked in here. But since the client is in a different project that is difficult to orchestrate.

So I am a bit confused.

mock/Readme.md
apozsuse: Fixed, Thanks!

mock/Readme.md
apozsuse: fixed, Thx!

apozsuse:
The goal of this mash API mock is to enable the possibility to perform integration tests for systems that interact with mash via its API. The driver for this was to have the chance to test brewmaster in the safe side, without the risk of getting something wrong 'published/deleted/whatever' in live environment.

Mash mock was the first step to implement such integration environment for brewmaster, but as you sat it does not benefit development mash server side in any form.

May be we should create a new repo for such tools instead of including them in this one?

smarlowucf:
It does add a lot of artifacts to the repo. Maybe it's better suited to live in our tools collection repo?

PR #843, opened 2022-11-29
Fixes issue with concurrent mp_changes in EC2 affecting same resource

apozsuse:

When mash tries to create concurrently change_sets that affect the same resources, boto3 raises ResourceInUseException and we have to resubmit.

This change includes a waiter that allows the second change_set to wait for the first to be completed and tries to apply the change_set again. 

Unit tests are included to provide 100% code coverage.
The new branch is only run in case ResourceInUseException is raised by boto3

### Additional Information

- Found out about `botocore.stub` to mock aws responses, I'll try to improve test coverage in some other repos...
- Number of reattempts and period default values might require tunning, as I wasn't sure about typical mp change_set completion times...
- Found out that utils/ec2.py was raising `MashGCEUtilsException`, will fix in another PR.

### DIFF
--- a/mash/utils/ec2.py
+++ b/mash/utils/ec2.py
@@ -17,12 +17,17 @@
 #
 
 import json
+import re
+import time
 
 import boto3
 
 from contextlib import contextmanager, suppress
 from mash.utils.mash_utils import generate_name, get_key_from_file
-from mash.mash_exceptions import MashGCEUtilsException
+from mash.mash_exceptions import (
+    MashEc2UtilsException,
+    MashGCEUtilsException
+)
 
 from ec2imgutils.ec2setup import EC2Setup
 from ec2imgutils.ec2removeimg import EC2RemoveImage
@@ -225,8 +230,21 @@ def start_mp_change_set(
     os_version,
     usage_instructions,
     recommended_instance_type,
-    ssh_user
+    ssh_user,
+    max_ResInUseExc_rechecks=20,
+    ResInUseExc_rechecks_period=300
 ):
+    """
+    Additional params included in this function:
+    - max_ResInUseExc_rechecks is the maximum number of checks that are
+    performed when a marketplace change cannot be applied because some resource
+    is affected by some other ongoing change (and ResourceInUseException is
+    raised by boto3).
+    - ResInUseExc_rechecks_period is the period (in seconds) that is waited
+    between checks for the ongoing mp change to be finished.
+    The maximum time the function will wait for a changeset to finish is
+    max_ResInUseExc_rechecks x ResInUseExc_rechecks_period seconds.
+    """
     data = {
         'ChangeType': 'AddDeliveryOptions',
         'Entity': {
@@ -265,9 +283,67 @@ def start_mp_change_set(
 
     data['Details'] = json.dumps(details)
 
-    response = client.start_change_set(
-        Catalog='AWSMarketplace',
-        ChangeSet=[data]
+    retries = 3
+    while retries > 0:
+        try:
+            response = client.start_change_set(
+                Catalog='AWSMarketplace',
+                ChangeSet=[data]
+            )
+            return response
+        except client.exceptions.ResourceInUseException as e:
+            # As Dec2022 there are no waiters in boto3 for marketplace-catalog
+            # client, implementing our own
+            wait_for_ongoing_change_to_complete(
+                get_ongoing_change_id_from_exc(
+                    e.response['Error']['Message']
+                ),
+                'AWSMarketplace',
+                client,
+                max_rechecks=max_ResInUseExc_rechecks,
+                rechecks_period=ResInUseExc_rechecks_period
+            )
+            retries -= 1
+        except Exception:
+            raise
+
+    raise MashEc2UtilsException(
+        f'Unable to submit the mp change for {ami_id}.'
     )
 
-    return response
+
+def wait_for_ongoing_change_to_complete(
+    change_id,
+    catalog,
+    client,
+    max_rechecks=20,
+    rechecks_period=300
+):
+    while max_rechecks > 0:
+        try:
+            response = client.describe_change_set(
+                Catalog=catalog,
+                ChangeSetId=change_id
+            )
+            # 'Status':'PREPARING'|'APPLYING'|'SUCCEEDED'|'CANCELLED'|'FAILED'
+            if all([
+                response,
+                'Status' in response,
+                response['Status'].lower().endswith('ed')
+            ]):
+                return
+        except Exception:
+            raise
+
+        time.sleep(rechecks_period)
+        max_rechecks -= 1
+
+
+def get_ongoing_change_id_from_exc(message: str) -> str:
+    change_id = ''
+    re_change_id = r'change sets: (\w+)[.,]'
+    match = re.search(re_change_id, message)
+
+    if match:
+        change_id = match.group(1)
+    return change_id
--- a/mash/mash_exceptions.py
+++ b/mash/mash_exceptions.py
@@ -189,3 +189,9 @@ class MashGCEUtilsException(MashException):
     """
     Exception raised if an error occurs in GCE Utils.
     """
+
+
+class MashEc2UtilsException(MashException):
+    """
+    Exception raised if an error occurs in EC2 Utils.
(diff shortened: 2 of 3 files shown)


mash/utils/ec2.py
```diff
@@ -17,12 +17,17 @@
...
+import time
 
 import boto3
 
 from contextlib import contextmanager, suppress
 from mash.utils.mash_utils import generate_name, get_key_from_file
-from mash.mash_exceptions import MashGCEUtilsException
+from mash.mash_exceptions import (
+    MashEc2UtilsException,
+    MashGCEUtilsException
```
rjschwei, line 29: We should not need GCEUtils in the EC2 implementation. Maybe a generic `UtilsException` would be suitable?

apozsuse: I think a generic utilsException would be suitable.
I did not address that change in this PR not to mix the scope, but I'll address that once this is closed/merged.

smarlowucf: This was likely a copy paste mistake.

mash/utils/ec2.py
```diff
@@ -225,8 +230,21 @@ def start_mp_change_set(
     os_version,
     usage_instructions,
     recommended_instance_type,
-    ssh_user
+    ssh_user,
+    max_ResInUseExc_rechecks=20,
```
smarlowucf, line 234: Snakecase. IMHO it could simply be max_rechecks and max_recheck_period. Also, the MP can take a few hours and that would be considered normal. So IMHO the period would be closer to 30 mins and bump the number of checks accordingly to keep total wait time the same.

mash/utils/ec2.py
```diff
@@ -265,9 +283,67 @@ def start_mp_change_set(
...
-        ChangeSet=[data]
+    retries = 3
+    while retries > 0:
+        try:
+            response = client.start_change_set(
+                Catalog='AWSMarketplace',
+                ChangeSet=[data]
+            )
+            return response
+        except client.exceptions.ResourceInUseException as e:
```
smarlowucf, line 294: Nitpick: s/e/error

mash/utils/ec2.py
```diff
@@ -265,9 +283,67 @@ def start_mp_change_set(
...
+    catalog,
+    client,
+    max_rechecks=20,
+    rechecks_period=300
+):
+    while max_rechecks > 0:
+        try:
+            response = client.describe_change_set(
+                Catalog=catalog,
+                ChangeSetId=change_id
```
smarlowucf, line 326: It's possible we get an empty string from get_ongoing_change_id_from_exc if the wording changes and breaks regex. Not sure how intricate we should get in error handling...

smarlowucf (commented):
I think it's good in general but I wonder if we want to have better error case handling. For example we hardcode the keys ['Error']['Message'] inside an exception case already. If that dict changes at some point it could lead to a very confusing keyerror being thrown while trying to handle a different exception path. Ditto if we can't find the changeset id in the error message or if the changeset cannot be found (I think this can be the case if it's a changeset created in the dashboard.)

That all being said there's a point where we go too far. Maybe there's a better way to organize the code such that we can make it less complex and limit the chances of an unhandled exception happening inside another exception clause?

While it might not be the pythonic route we may want to check existing changesets before attempting to create a new one. This adds the window for a race condition if jobs are running at the same time but the window should be pretty small. E.g.

- Check for any existing changesets that are not in a finalized state
- If none are running then create a new change set
- If one is running then what and check later
- If we wait too long then raise and exit

Another option is to just keep trying to add the changeset. If we get resource in use then we wait longer and try again. If we get some other error we raise and exit. If we wait too long then raise and exit.

rjschwei:
> I think it's good in general but I wonder if we want to have better error case handling. For example we hardcode the keys ['Error']['Message'] inside an exception case already. If that dict changes at some point it could lead to a very confusing keyerror being thrown while trying to handle a different exception path. 

The error is part of the public API. unlike others AWS has a strong record of not breaking their public API. As such I don;t think this would be an issue we encounter in practice.


> Ditto if we can't find the changeset id in the error message or if the changeset cannot be found (I think this can be the case if it's a changeset created in the dashboard.)

I am not certain I follow you. Are you eluding to a possible collision between a changeset (request created in the UI) and a changeset created through the update process? In that case shouldn't the error, and of course we have to ask or find documentation, testing this is not going to work, information returned give us the changeset id that is already posted and blocking this submission?

smarlowucf:
> Are you eluding to a possible collision between a changeset (request created in the UI) and a changeset created through the update process? In that case shouldn't the error, and of course we have to ask or find documentation, testing this is not going to work, information returned give us the changeset id that is already posted and blocking this submission?

Not sure I follow the question exactly. If you start a change in the UI it creates a pending changeset. We tripped over this recently and when I searched for the changeset in cli I got nothing back. So either the changeset was older than 60 days or was not visible. Either way you cannot have two changesets at once. This is an issue for multi sku offers.

rjschwei:
> > Are you eluding to a possible collision between a changeset (request created in the UI) and a changeset created through the update process? In that case shouldn't the error, and of course we have to ask or find documentation, testing this is not going to work, information returned give us the changeset id that is already posted and blocking this submission?
> 
> Not sure I follow the question exactly. If you start a change in the UI it creates a pending changeset. We tripped over this recently and when I searched for the changeset in cli I got nothing back. So either the changeset was older than 60 days or was not visible. Either way you cannot have two changesets at once. This is an issue for multi sku offers.

Ok that kind of answers my question. I would have expected a "changeset blocked by ....$SOMEID" response if the changeset that is being submitted is not accepted. Sounds like my expectations are out of whack with the real world. With this in mind it looks like the "re-try submission until some total time out is reached" is the only approach we can take.

smarlowucf:
No that is correct. We get a response with the blocking changeset ids. That's how it's implemented currently. We get that and then wait for the changeset to finish to move on. But we are dropping into nested try/except clauses which I think will make any error cases difficult to debug.

rjschwei:
> No that is correct. We get a response with the blocking changeset ids. That's how it's implemented currently. We get that and then wait for the changeset to finish to move on. But we are dropping into nested try/except clauses which I think will make any error cases difficult to debug.

Well the nested try/except is a matter of refactoring into methods.

mash/utils/ec2.py
apozsuse: Done!
Simplified the names for the params and adjusted the values to wait 5 hours max for a changeset to finish.

apozsuse:
Refactored the error handling part when waiting for the conflicting changeset to be finished.

Any unexpected error in the waiting part raises an exception now...

mash/utils/ec2.py
apozsuse: As the `MashGCEUtilsException` was just used in one place in `ec2_utils` file, I included the fix in this PR.

mash/utils/ec2.py
```diff
@@ -265,9 +278,86 @@ def start_mp_change_set(
...
+            # As Dec2022 there are no waiters in boto3 for marketplace-catalog
+            # client, implementing our own
+            try:
+                ongoing_change_id = get_ongoing_change_id_from_error(
+                    conflicting_error_message
+                )
+                wait_for_ongoing_change_to_be_completed(
+                    ongoing_change_id,
+                    'AWSMarketplace',
+                    client,
```
smarlowucf, line 310: Sorry didn't notice this before: This will be a long running check so there's a good chance the client session expires. Given client is injected we may want error handling in calling scope instead.

smarlowucf (commented):
Sorry, missed the issue of session expiry. We'll have to work around that somehow.

mash/utils/ec2.py
rjschwei: Agreed when we get into the error handling loop where we wait for a changeset publish to complete the client will definitely expire.

apozsuse: Refactored the solution a bit to allow the code to generate a new client when waiting for the conflicting change set to finish.
The API for some function calls is changed to allow that (boto3 dependency injection is removed and boto3 auth data is included). We will have to bump a major version when creating a release...

smarlowucf (approved):
LGTM

rjschwei (approved):

PR #848, opened 2023-02-27
Modifies s3bucket upload service to allow raw image upload to a S3 bucket

apozsuse:

We need to have to have the possibility of just uploading raw images to a s3 bucket (without image publication)

These changes:
- Allows to use the s3upload service without `cloud_image_name` parameter (and without image publication)
- Includes the option to `use_build_time` in the `raw_image_upload_location` parameter

The changes have been tested.

Usual procedure without special considerations.

### Additional Information
 None

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -7,7 +7,7 @@
 
 Flask
 flask-restx
-flask-sqlalchemy<=3.0.0
+flask-sqlalchemy<3.0.0
 flask-migrate
 flask-jwt-extended>=4.0.0
 setuptools>=40.3.0
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -29,3 +29,4 @@ google-cloud-storage
 google-api-python-client
 aliyun-img-utils>=1.4.0
 azure-img-utils>=0.4.1
+Sqlalchemy<2.0.0
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -7,7 +7,7 @@
 
 Flask
 flask-restx
-flask-sqlalchemy
+flask-sqlalchemy=2.4.4
 flask-migrate
 flask-jwt-extended>=4.0.0
 setuptools>=40.3.0
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -7,7 +7,7 @@
 
 Flask
 flask-restx
-flask-sqlalchemy<=2.4.4
+flask-sqlalchemy<=3.0.0
 flask-migrate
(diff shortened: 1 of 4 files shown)


rjschwei (approved):

.virtualenv.requirements.txt
```diff
@@ -7,7 +7,7 @@
 
 Flask
 flask-restx
-flask-sqlalchemy
```
smarlowucf, line 10: We can't lock into a specific version as that breaks downstream packaging. Requirements should be >= or <=.

smarlowucf: Most likely we need the <=.

mash/services/upload/s3bucket_job.py
```diff
@@ -79,9 +91,15 @@ def run_job(self):
         except IndexError:
             bucket_path = ''
 
-        if self.status_msg['cloud_image_name']:
+        if (
+            'cloud_image_name' in self.status_msg and self.status_msg[
```
smarlowucf, line 95: could be simplified to  `if self.status_msg.get('cloud_image_name')`

smarlowucf (changes requested):

.virtualenv.requirements.txt
apozsuse: OK! Had to modify the setting because the CI was failing due to an incompatibility in flask-sqlalchemy in 3.0.2
Fixed!

mash/services/upload/s3bucket_job.py
apozsuse: There are some cases (at least in the tests) where 'cloud_image_name' is present in the dict but it has None value :S
Probably it's something just in the tests, but just in case I added the second check to validate the value is not None.

.virtualenv.requirements.txt
smarlowucf: If the breaking change is in 3.0.2 then I guess it probably came in 3.0.0 thus would put the conditional as < 3.0.0.

mash/services/upload/s3bucket_job.py
apozsuse: Fixed!

apozsuse:
About the issue with the requirements.txt, finally realized that the breaking change happens in SQLAlchemy 2.0.0, so I included an explicit requirement to have a version < 2.0.0

smarlowucf (approved):
I think we only need one of the requirement pins.

.virtualenv.requirements.txt
```diff
@@ -7,7 +7,7 @@
...
+flask-sqlalchemy<3.0.0
```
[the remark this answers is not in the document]
apozsuse: Fixed!

Issue #854, opened 2023-08-29
Testing of specific instance types in specific regions


rjschwei:
AMD-SEV is a feature supported by all CSPs and it is reasonable to expect that at some point TDX will be in the feature set as well. mash should be able to drive img-proof such that we can test booting with AMS SEV-SNP and eventually Intel TDX instances. This is in addition to the already existing capabilities of `bios` vs. `uefi`. This is related to https://github.com/SUSE-Enceladus/img-proof/issues/363

PR #855, opened 2023-09-19
Encapsulating 'test_image' call into a function for EC2 image tests

apozsuse:
 This is a first PR to address [issue #854](https://github.com/SUSE-Enceladus/mash/issues/854) and be able to run specific tests per region with different instance flags.

There are no functional changes in this PR. It just encapsulates  the `test_image` call where the tests are run into a function so it's easier to add specific tests for some regions (where some features might be supported)  

Unit tests have been run to the changed code.

Nope.

### Additional Information

My idea is to add a new field to the jobdoc where we will be able to overwrite some fields for the tests (region, cpuOptions, etc.)

### DIFF
--- a/mash/services/test/ec2_job.py
+++ b/mash/services/test/ec2_job.py
@@ -145,70 +145,31 @@ class EC2TestJob(MashJob):
                     # There are no aarch64 based instance types available.
                     continue
 
-                with setup_ec2_networking(
-                    credentials['access_key_id'],
-                    region,
-                    credentials['secret_access_key'],
-                    self.ssh_private_key_file,
-                    subnet_id=info.get('subnet')
-                ) as network_details:
-                    try:
-                        exit_status, result = test_image(
-                            self.cloud,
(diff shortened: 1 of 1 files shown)


mash/services/test/ec2_job.py
```diff
@@ -221,3 +182,79 @@ def run_job(self):
...
+        description,
+        distro,
+        image_id,
+        instance_type,
+        img_proof_timeout,
+        ssh_user,
+        tests,
+        log_callback
+    ):
+        with setup_ec2_networking(
```
smarlowucf, line 203: A multi test scenario would happen in the same region/zone. So we should be able to use the same ec2 networking.

smarlowucf (approved):

mash/services/test/ec2_job.py
apozsuse: I hadn't thought about that. Makes sense.
I changed the network creation out of the run_tests fn. Thx!

PR #856, opened 2023-09-22
Small refactor in test service for ec2 to include the possibility of SEV testing

apozsuse:
Could not finish this code change. I'll retake the job after the workshop...

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -19,7 +19,7 @@ APScheduler
 python-dateutil>=2.6.0
 amqpstorm
 ec2imgutils>=9.0.3
-img-proof>=7.14.0
+img-proof>=7.21.0
 requests
 obs-img-utils>=1.0.0
 oci
--- a/mash/services/test/ec2_job.py
+++ b/mash/services/test/ec2_job.py
@@ -42,15 +42,23 @@ instance_types = {
         'm5.large',
         't3.small'
     ],
+    'hybrid-sev': [
+    ],
     'bios': [
         'i3.large',
         't2.small'
+    ],
+    'bios-sev': [
     ],
     'aarch64': [
         't4g.small',
         'm6g.medium'
     ]
 }
+
+sev_capable_regions = [
+    'us-east-1'
+]
 
 
 class EC2TestJob(MashJob):
@@ -86,27 +94,12 @@ class EC2TestJob(MashJob):
             ['uefi-preferred']
         )[0]
 
-        if self.cloud_architecture == 'aarch64':
-            self.instance_types = [
-                random.choice(instance_types['aarch64'])
-            ]
-        elif self.boot_firmware == 'uefi-preferred':
-            self.instance_types = [
-                random.choice(instance_types['bios']),
-                random.choice(instance_types['hybrid'])
-            ]
-        elif not self.instance_type:
-            self.instance_types = [
-                random.choice(instance_types['hybrid'])
-            ]
-        else:
-            self.instance_types = [self.instance_type]
-
         self.ssh_private_key_file = self.config.get_ssh_private_key_file()
         self.img_proof_timeout = self.config.get_img_proof_timeout()
 
         if not os.path.exists(self.ssh_private_key_file):
             create_ssh_key_pair(self.ssh_private_key_file)
+        self.guest_os_features = self.job_config.get('guest_os_features', [])
 
     def run_job(self):
         """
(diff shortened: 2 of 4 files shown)


.virtualenv.requirements.txt
```diff
@@ -19,7 +19,7 @@ APScheduler
 python-dateutil>=2.6.0
 amqpstorm
 ec2imgutils>=9.0.3
-img-proof>=7.14.0
+img-proof>=7.21.0
```
apozsuse, line 22: Included this change to assure `img-proof` has the new param.

mash/services/test/ec2_job.py
```diff
@@ -42,16 +42,24 @@
         'm5.large',
         't3.small'
     ],
+    'hybrid-sev': [
+    ],
```
apozsuse, line 46: Pending.
We'd need to check which instance_types we want to test for SEV_CAPABLE images when UEFI is supported.

apozsuse:
This one is ready for review.

mash/services/test/ec2_job.py
```diff
@@ -127,31 +122,49 @@ def run_job(self):
...
+
+            if info['partition'] in ('aws-cn', 'aws-us-gov') and \
+                    self.cloud_architecture == 'aarch64':
+                # Skip test aarch64 images in China and GovCloud.
+                # There are no aarch64 based instance types available.
+                continue
+
+            if all([
+                info['partition'] == 'aws',
+                'SEV_CAPABLE' in self.guest_os_features,
```
rjschwei, line 135: 'SEV_CAPABLE' is a Google setting, it is a bit confusing to see it pop up in the EC2 code stream.  Is the plan to set this in the job docs for EC2? And how is this hen documented in a sensible way for other users (not us) of mash?

apozsuse: Yes, the plan is to set those settings in the job docs where suitable.
I reused the Google setting names but I'm ok with changing these settings to match AWS terminology to make them more explicit:
  - I'll change 'guest_os_features' in the job doc for 'cpu_options'
  - 'SEV_CAPABLE' for 'AMD_SEV_SNP_ENABLED' 
  
WDYT?

smarlowucf: Yeah, I agree on using aws specific terminology here.

mash/services/test/ec2_job.py
```diff
@@ -52,6 +57,10 @@
     ]
 }
 
+sev_capable_regions = [
+    'us-east-1'
```
smarlowucf, line 61: The feature is only available in us-east-2 and eu-west-1 regions.

smarlowucf (changes requested):

smarlowucf (approved):

mash/services/test/ec2_job.py
```diff
@@ -127,31 +123,49 @@ def run_job(self):
...
+                # There are no aarch64 based instance types available.
+                continue
+
+            if all([
+                info['partition'] == 'aws',
+                'AMD_SEV_SNP_ENABLED' in self.cpu_options,
+                region in sev_snp_supported_regions,
+                self.cloud_architecture == 'x86_64'
+            ]):
+                self.sev_snp_enabled = True
```
rjschwei, line 140: Come to think of it I do not think this should be a flag, i.e. true/false. We should expect TDX and we will have to test TDX instances as well. What if we change the name of the variable to `mem_encryption` with a value of  `SEV` , `TDX`, or  `NONE` and then the `select_instance_type` method can switch on that setting and either pick from `hybrid-sev-snp` (I think that could maybe simply be `sev-snp-instances`, not sure why `hybrid` is supposed to express) or the future `tdx-instances`. Then the condition can then be expressed as:
```
self.mem_encryption = 'NONE'
if info['partition'] == 'aws' and self.cloud_architecture == 'x86_64':
    if 'AMD_SEV_SNP_ENABLED' in self.cpu_options and region in sev_snp_supported_regions:
        self.mem_encryption = 'SEV'
    if 'INTEL_TDX_ENABLED' in self.cpu_options and region in tdx_supported_regions:
        self.mem_encryption = 'TDX'
```
And of course this really applies to all CSPs as such this should really be lifted to the base class of the jobs. A bit more abstraction is needed for that. It's a bit bigger change but we need the same logic for all 3 CSPs. The requirement can be boiled down to selecting a specific instance type in a specific region when either TDX or SEV is set and run that instance with a specific flag. This is common across all 3 CSPs we support.

rjschwei (commented):
I am not opposed to merging this and coming back with a follow up PR. That said, we are not in a hurry, SEV-SNP in AWS is currently broken and we need to get that fixed before we enable this testing.

apozsuse:
Closing this PR as this will be implemented in the test refactor.

PR #857, opened 2023-10-11
Use upstream name filter in describe images call

smarlowucf:
Raise exception if more than one image is found.

Note: This is pre-work for enabling version restriction in aws mp.


rjschwei (approved):

apozsuse (approved):
LGTM

PR #860, opened 2023-10-12
Add function to retrieve delivery option id

smarlowucf:
For the version based on provided ami id and entity id.


smarlowucf:
Note: jmespath was already an implicit dependency coming from boto3. Adding it to requirements due to direct usage but it does not affect deployment.

smarlowucf:
Hold off on review. This is more complex...

smarlowucf:
Okay untangled the data structure here. Ready for review.

rjschwei (approved):

PR #863, opened 2023-10-13
Add filters option to describe images function

smarlowucf:
This provides the option to filter images on any attr including image name.


apozsuse (approved):
LGTM

PR #869, opened 2023-12-15
Add launch_inst_type option for ec2 jobs

smarlowucf:
This provides a way to override the default instance type to use in ec2uploadimg for ec2 image creation.


apozsuse (approved):
LGTM :+1:

PR #870, opened 2023-12-15
Enable version restriction for aws mp jobs

smarlowucf:
Version restriction is triggered based on the old image name attr. If it exists the restriction happens in deprecate service. The change set is submitted after the last change is created in either publish or deprecate service.

### DIFF
--- a/mash/services/deprecate/ec2_mp_job.py
+++ b/mash/services/deprecate/ec2_mp_job.py
@@ -30,7 +30,7 @@ from mash.utils.ec2 import (
 
 class EC2MPDeprecateJob(MashJob):
     """
-    Class for an EC2 deprecate job.
+    Class for EC2 Marketplace restriction.
     """
 
     def post_init(self):
@@ -54,7 +54,7 @@ class EC2MPDeprecateJob(MashJob):
 
     def run_job(self):
         """
-        Deprecate image in all target regions in each source region.
+        Restrict the image in each source region.
         """
         self.status = SUCCESS
 
--- a/mash/services/deprecate/ec2_mp_job.py
+++ b/mash/services/deprecate/ec2_mp_job.py
@@ -1,0 +1,104 @@
+# Copyright (c) 2023 SUSE LLC.  All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


mash/services/deprecate/ec2_mp_job.py
```diff
@@ -0,0 +1,104 @@
...
+                )
+            )
+
+        self.old_cloud_image_name = self.job_config.get(
+            'old_cloud_image_name'
+        )
+
+    def run_job(self):
+        """
+        Deprecate image in all target regions in each source region.
```
apozsuse, line 57: s/in each source region//

mash/services/publish/ec2_mp_job.py
```diff
@@ -101,23 +105,28 @@ def run_job(self):
...
+
+                self.status_msg['change_set_id'] = response.get('ChangeSetId')
+                self.log_callback.info(
+                    'Marketplace change set submitted. Change set id: '
+                    '{change_set}'.format(
+                        change_set=self.status_msg['change_set_id']
+                    )
+                )
+            else:
+                self.status_msg['add_version_doc'] = change_doc
```
apozsuse, line 129: I don't fully understand how, in the case we've got more than one mp account to publish, we're not overwriting the `add_version_doc` key in the `status_msg` dictionary.

I see these are consumed (`pop` ed) in the deprecate service, but does this mechanism assure the messages are consumed by the deprecate service before being overwritten by the publish service again? :thinking:

rjschwei (approved):

mash/services/deprecate/ec2_mp_job.py
smarlowucf: Updated the message.

mash/services/publish/ec2_mp_job.py
smarlowucf: This is a known bug in the MP path. The same issue exists in other MP specific services. Thus I think it's best to handle this all in one PR. Created an issue to track of this https://github.com/SUSE-Enceladus/mash/issues/872.

smarlowucf:
Will hold this open until after the holidays in case there's a need to push a new mash version in the meantime. This way we limit changeset in that case.

PR #871, opened 2023-12-18
Increase length of download url

### LINKED ISSUE #864: Download url too short
Download URL is capped at 128. This is easily surpassed when a project is forked to home project. This prevents the project from being used as it causes an error when the job is saved to db.

smarlowucf:
Branched projects can end up with very long urls.

### Additional Information

Fixes #864


apozsuse (approved):
LGTM

rjschwei (approved):

PR #873, opened 2023-12-20
Renames the service that gets the images from `obs` service to `download` service

apozsuse:
First PR from a series to make the download service extendable. This will allow us to have different image sources than IBS repositories (as for example S3 buckets).

This PR just changes the name of the service from `obs` to a more generic `download` service.

Note that in order not to break `mash` build, we're using a feature branch that will be merged into main when the whole change is finished and tested.

### DIFF
--- a/mash/services/download/obs_job.py
+++ b/mash/services/download/obs_job.py
@@ -21,7 +21,7 @@ import dateutil.parser
 
 # project
 from mash.services.mash_service import MashService
-from mash.services.obs.build_result import OBSImageBuildResult
+from mash.services.download.obs_build_result import OBSImageBuildResult
 from mash.utils.json_format import JsonFormat
 from mash.utils.mash_utils import persist_json, restart_jobs, setup_logfile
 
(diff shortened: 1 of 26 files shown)


rjschwei (approved):

mash/services/download/obs_job.py
smarlowucf, line 1: This will probably stay as a generic service file for the download service. The build_result module is a closer approximation to the obs job class.

smarlowucf (commented):
Looks good, I would leave service.py file as it is and rename `build_result` to `obs_job`.

mash/services/download/obs_job.py
apozsuse: I had a look at the other services and most of them don't have `generic` service files under `mash/services/whatever` directory, but they implement a `BaseJobFactory` (in a file under `mash/services`  to select the specific subclass in the service subdirectory.

I'll merge this PR and address this name files issue in the next PR when the BaseJobFactory is included, ok?

smarlowucf: Yeah, those services are all drive a bit differently with 2 separate messages.

PR #874, opened 2023-12-21
Remove some last obs service references

apozsuse:
Found some missing references to `obs` service that where not changed in the previous PR, like the Readme file and some tests.


rjschwei (commented):
LGTM

PR #876, opened 2024-01-04
S3 download class implementation

apozsuse:
This PR includes the implementation for the `S3DownloadJob` class.

### File naming scheme

The naming scheme agreed for the image files in the bucket is the following:

```
AWS:
    suse-liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.raw.xz

Azure
    liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.vhdfixed.xz

Google
    liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.tar.gz
```

A couple new fields in the `mash` job document schema are included in this PR to keep mash abstracted from these details:

- `s3_download_file_prefix`: A prefix to be prepended to the image_name for the file download. The expected value for this is `suse-` for AWS and empty string (or not include the field) for the rest of cloud providers.
- `s3_download_file_suffix`: Same but for the suffix to be appended after the `$ARCH` segment.

For the bucket address the `download_url` file is reused.

### Authentication

A new field has been included(`download_account`) in the `mash` job document to include the account that will be used to authenticate in aws to access the S3 bucket.
The credentials corresponding to that account have to be provisioned in mash as usual.

### DIFF
--- a/mash/services/api/v1/schema/jobs/__init__.py
+++ b/mash/services/api/v1/schema/jobs/__init__.py
@@ -92,6 +92,19 @@ base_job_message = {
                         'Build1.206.vhdfixed.xz.sha256 The "image" is '
                         'openSUSE-Leap-15.0-EC2-HVM.'
         ),
+        's3_download_file_prefix': string_with_example(
+            'suse-',
+            description='This should include any prefix that is prepended in'
+                        'the image filename present when image files are'
+                        'downloaded from an S3 bucket'
+        ),
+        's3_download_file_suffix': string_with_example(
+            '.tar.gz',
+            description='This should include any suffix that is included in '
+                        'the image filename present, after the architecture'
+                        'substring when image files are downloaded from an S3'
+                        ' bucket.'
+        ),
         'cloud_image_name': string_with_example(
             'openSUSE-Leap-15.0-v{date}-hvm-ssd-x86_64',
             description='The name to use for the uploaded image in the cloud '
@@ -128,6 +141,14 @@ base_job_message = {
                            'itself. Valid condition operators are >, <, >=, '
                            '<=, or ==.'
         },
+        'download_account': string_with_example(
+            'my_aws_account',
+            description='The cloud framework account as configured with '
+                        'the mash account add client command when the mash '
+                        'user was setup. The credentials associated with '
+                        'this cloud framework account will be used for the '
+                        'image download.'
+        ),
         'download_type': string_with_example(
             'OBS',
             description='Service from which to download the image file.'
@@ -140,9 +161,12 @@ base_job_message = {
         'download_url': string_with_example(
             'https://download.opensuse.org/repositories/'
             'Cloud:/Images:/Leap_15.0/images/',
-            description='The URL to a download repository. The URL is '
+            description='The URL to a download repository.'
+                        'In the case of OBS download, the URL is '
                         'expected to have the image tarball, checksum and '
                         'a packages file.'
+                        'For the S3 download case, it contains the URL of the'
+                        'S3 bucket containing the image file.'
         ),
         'image_description': string_with_example(
             'openSUSE Leap 15.0 (HVM, 64-bit, SSD-Backed)',
--- a/mash/services/download/s3_job.py
+++ b/mash/services/download/s3_job.py
@@ -21,8 +21,7 @@ import logging
 import re
 import os
 
-from datetime import datetime
-from zoneinfo import ZoneInfo
+from datetime import datetime, timezone
 from pytz import utc
 from apscheduler.schedulers.background import BackgroundScheduler
 from apscheduler.events import EVENT_JOB_SUBMITTED
@@ -212,7 +211,7 @@ class S3DownloadJob(object):
             date_str = match.group(1)
             date_str = date_str + 'T00:00:00.000'
             date = datetime.strptime(date_str, '%Y%m%dT%H:%M:%S.%f').replace(
-                tzinfo=ZoneInfo('UTC')
+                tzinfo=timezone.utc
             )
             return date.strftime('%s')
         else:
--- a/mash/services/download/s3_job.py
+++ b/mash/services/download/s3_job.py
@@ -18,10 +18,9 @@
 
 
 import logging
-import re
 import os
 
-from datetime import datetime, timezone
+from datetime import datetime
 from pytz import utc
 from apscheduler.schedulers.background import BackgroundScheduler
 from apscheduler.events import EVENT_JOB_SUBMITTED
@@ -206,16 +205,7 @@ class S3DownloadJob(object):
             )
 
     def _get_build_time(self, image_name):
-        match = re.search(r'^.*-v(\d{8})-.*', image_name)
-        if match:
-            date_str = match.group(1)
-            date_str = date_str + 'T00:00:00.000'
-            date = datetime.strptime(date_str, '%Y%m%dT%H:%M:%S.%f').replace(
-                tzinfo=timezone.utc
-            )
-            return date.strftime('%s')
-        else:
-            return 'unknown'
+        return 'unknown'
 
     def _job_skipped_event(self, event):
         # Job is still active while the next _update_image_status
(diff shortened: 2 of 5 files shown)


smarlowucf (commented):
If there was an overlap with extensions between two clouds and images were uploaded on the same day the files would stomp on each other. Should we request that files are uploaded using csp names that are included in the filename like:

```
google/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.tar.gz
microsoft/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.vhdfixed.xz
amazon/suse-liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.raw.xz
```

Where the prefix could be the "cspname/..." part.

rjschwei:
> If there was an overlap with extensions between two clouds and images were uploaded on the same day the files would stomp on each other. Should we request that files are uploaded using csp names that are included in the filename like:
> 
> ```
> google/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.tar.gz
> microsoft/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.vhdfixed.xz
> amazon/suse-liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.raw.xz
> ```
> 
> Where the prefix could be the "cspname/..." part. Maybe this isn't an issue now but something to consider for the future. I believe OCI uses qcow2 files so there's no current overlap.

There should never be an overlap

> If there was an overlap with extensions between two clouds and images were uploaded on the same day the files would stomp on each other. Should we request that files are uploaded using csp names that are included in the filename like:
> 
> ```
> google/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.tar.gz
> microsoft/liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.vhdfixed.xz
> amazon/suse-liberty-$MAJOR-$MINOR-byos-vYYYYMMDD-$ARCH.raw.xz
> ```
> 
> Where the prefix could be the "cspname/..." part. Maybe this isn't an issue now but something to consider for the future. I believe OCI uses qcow2 files so there's no current overlap.

Well the only possible overlap would be if some other cloud framework implementation expected a tarball that is gzip compressed and the format of the data in the image would be different than what we expect for GCE. That said the separation into subdirectories in general, as suggested, to avoid potential collisions is already supported via the `download_url` setting in the jobdoc .

rjschwei:
I am not convinced that we need `s3_download_file_suffix` and `s3_download_file_prefix`.

For our use case where we want to support image uploads for images build outside the Open Build Service and pull those images from an S3 bucket we simply prescribe that the name prior to the first `.` must match our image names as they appear in the frameworks where we publish images. For all other users the construction of the filename via `s3_download_file_suffix` and `s3_download_file_prefix` is probably  cumbersome.

Secondly `mash` makes assumptions about how the image file format relates to a specific framework, for example one cannot create an image in EC2 with mash if the origin build is in qcow2 format. Therefore I think sticking to the approach taken in obs-img-utils [1] is the way to go here as well. Loop over the defined extensions and prepend each with the known `image-name` and that's what we download.

[1] https://github.com/SUSE-Enceladus/obs-img-utils/blob/main/obs_img_utils/utils.py#L36

apozsuse:
> Secondly `mash` makes assumptions about how the image file format relates to a specific framework, for example one cannot create an image in EC2 with mash if the origin build is in qcow2 format. Therefore I think sticking to the approach taken in obs-img-utils [1] is the way to go here as well. Loop over the defined extensions and prepend each with the known `image-name` and that's what we download.
> 
> [1] https://github.com/SUSE-Enceladus/obs-img-utils/blob/main/obs_img_utils/utils.py#L36

The `obs-img-utils` approach works in OBS because we've got the `profile` field that makes the image name (without the extension) `unique` per cloud (the name is unique, so we just loop around the possible extensions until we find one matching).

In this case, as the name can be equal for 2 different clouds (if we need to publish amazon and google the same day for example), we need to include the logics to get the exact file name (with the extension) depending on the cloud naming schema somewhere.

May be having a single `download_file_name` field in the mash job doc would be easier than having a prefix and suffix?

smarlowucf:
Agree on the prefix. The "suse-" part for ec2 images is treated as part of the image name everywhere else. So an extra suffix arg isn't really needed.

Based on the example filenames provided here they don't compare to what obs-img-utils is working with though. In that case the format is like `SLES15-SP5-BYOS.aarch64-1.0.3-GCE-Build1.9.tar.gz`. There is a token in there to distinguish cloud framework. Unless there's a token in the filenames in s3 with this data how will we know where the image should go?

rjschwei:
> Agree on the prefix. The "suse-" part for ec2 images is treated as part of the image name everywhere else. So an extra suffix arg isn't really needed.
> 
> Based on the example filenames provided here they don't compare to what obs-img-utils is working with though. In that case the format is like `SLES15-SP5-BYOS.aarch64-1.0.3-GCE-Build1.9.tar.gz`. There is a token in there to distinguish cloud framework. Unless there's a token in the filenames in s3 with this data how will we know where the image should go?

Well, we should not build the answer  of "where the image should go" based on the extension of the file name or a "token" in the filename and we have not done that. Nor should we build around whether or not there is only 1 file downloaded (more on the one file downloaded in the answer to @apozsuse in a minute). The jobdoc has the `provider` key and that answers where the image should go. So in general the schematics in obs-img-utils will still work.

rjschwei:
> > Secondly `mash` makes assumptions about how the image file format relates to a specific framework, for example one cannot create an image in EC2 with mash if the origin build is in qcow2 format. Therefore I think sticking to the approach taken in obs-img-utils [1] is the way to go here as well. Loop over the defined extensions and prepend each with the known `image-name` and that's what we download.
> > [1] https://github.com/SUSE-Enceladus/obs-img-utils/blob/main/obs_img_utils/utils.py#L36
> 
> The `obs-img-utils` approach works in OBS because we've got the `profile` field that makes the image name (without the extension) `unique` per cloud (the name is unique, so we just loop around the possible extensions until we find one matching).
> 
> In this case, as the name can be equal for 2 different clouds (if we need to publish amazon and google the same day for example), we need to include the logics to get the exact file name (with the extension) depending on the cloud naming schema somewhere.
> 
> May be having a single `download_file_name` field in the mash job doc would be easier than having a prefix and suffix?

Having a `download_file_name`  works for me. That said it is not strictly necessary. In the example provided where image files for different clouds are the same but for the extension we would simply download 2 files if we were to loop over the expected extensions. The mash uploader, for example [1] then finds the proper file for the cloud target specified with the `provider` key in the jobdoc.

That said in order to be resource consumption friendly on only download what we need at the time we need it having  a `download_file_name` key is the responsible approach.


[1] https://github.com/SUSE-Enceladus/mash/blob/main/mash/services/upload/gce_job.py#L87

smarlowucf:
Isn't the file showing up in the s3 bucket that triggers this process though? If that's the initial trigger then it needs to communicate all the necessary info in some way; filename, cloud, image name. There's many ways we can go about that. But if we have one bucket then one example could be filename of: `google/liberty-15-6-byos-v20240101-x86_64.tar.gz`. E.g. `{cloud}/{image name}.{ext}`. Another option is one bucket per cloud and filename then could be `{image name}.{ext}`. Either way the monitor script would parse this and could prevent any SUSE specific logic getting into mash.

rjschwei:
> Isn't the file showing up in the s3 bucket what triggers this process though? If that's the initial trigger then it needs to communicate all the necessary info in some way; filename, cloud, image name. There's many ways we can go about that. But if we have one bucket then one example could be filename of: `google/liberty-15-6-byos-v20240101-x86_64.tar.gz`. E.g. `{cloud}/{image name}.{ext}`. Another option is one bucket per cloud and filename then could be `{image name}.{ext}`. Either way the monitor script would parse this and could prevent any SUSE specific logic getting into mash.

This gets into a conversation that is outside the scope of this PR and some parts of that conversation should not be public.

apozsuse:
Closing this PR to create new ones with the agreed approach to this feature.

PR #879, opened 2024-01-10
Ec2 utils get bucket file list

apozsuse:
This PR includes a new function in the ec2 utils files that allows to get a list of the files in a S3 bucket.

The function includes the option to include a regex and get the file names in the bucket that are a `fullmatch` to the regex provided.

### DIFF
--- a/mash/utils/ec2.py
+++ b/mash/utils/ec2.py
@@ -457,13 +457,18 @@ def get_ongoing_change_id_from_error(message: str):
         )
 
 
-def get_file_list_from_s3_bucket(s3_client, bucket_name, filter_regex=None):
+def get_file_list_from_s3_bucket(
+    boto3_session,
+    bucket_name,
+    filter_regex=None
+):
     """Provides the list of files in a bucket
     If a regex is provided in filter_regex parameter, only the matching
     files will be included in the list returned.
     For the file to be included in the returned list, the S3 object name has to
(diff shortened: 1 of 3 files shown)


rjschwei (approved):

mash/utils/ec2.py
```diff
@@ -455,3 +455,28 @@ def get_ongoing_change_id_from_error(message: str):
         raise MashEc2UtilsException(
             f'Unable to extract changeset id from aws err response: {message}'
         )
+
+
+def get_file_list_from_s3_bucket(s3_client, bucket_name, filter_regex=None):
```
smarlowucf, line 460: Nitpick/suggestion: instead of requiring a client this could be a session arg. The advantage would be that the calling scope does not need to know the client details. Of course that advantage depends on what else the calling scope is doing. E.g.

```
def get_file_list_from_s3_bucket(boto3_session, bucket_name, filter_regex=None):
    s3_client = boto3_session.client('s3')
```

If calling scope is also using the client in some way then the suggestion is moot.

smarlowucf (approved):

mash/utils/ec2.py
apozsuse: Implemented this change, as the caller of the function can use a Session and all the calls to s3 clients should be included in utils functions.

PR #880, opened 2024-01-10
Implements a function to download files from a S3 bucket in ec2 utils file

apozsuse:


rjschwei (approved):

mash/utils/ec2.py
```diff
@@ -485,3 +486,20 @@ def get_file_list_from_s3_bucket(
...
+    bucket_name,
+    file_name,
+    download_directory
+):
+    """Downloads a file from a S3 bucket to the provided directory"""
+
+    if not os.path.exists(download_directory):
+        os.makedirs(download_directory)
+
+    file_full_path = os.path.join(download_directory, file_name)
```
smarlowucf, line 502: nitpick: file_full_path could be shortened to file_path or download_path. Currently it could read as file-full path which would be confusing.

smarlowucf (approved):

mash/utils/ec2.py
apozsuse: I changed it to download_path for better readability. Thx!

PR #881, opened 2024-01-11
Implements a function to get the regex to filter all the images for the image_name and cloud provided

apozsuse:
This function provides a regex that matches the different versions for the same image taking into account the cloud provider:
- It removes the first '-' delimited segment in the name for some cloud vendors.
- Postpones the appropriate file extension for the cloud provider.

### DIFF
--- a/mash/services/download/s3bucket_job.py
+++ b/mash/services/download/s3bucket_job.py
@@ -19,6 +19,7 @@
 
 import logging
 import os
+import re
 
 from datetime import datetime
 from pytz import utc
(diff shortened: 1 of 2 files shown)


mash/services/download/s3bucket_job.py
```diff
@@ -180,3 +181,43 @@ def _job_skipped_event(self, event):
         # event was scheduled. In this case we just skip the event
         # and keep the active job waiting for an obs change
         pass
+
+    def _get_regex_for_filename(self, image_name, cloud):
```
rjschwei, line 185: This is very SUSE specific and we should not do this in mash. Handling the extensions is one thing, pretty much anyone else uploading images will need image files in the same formats we use. However, this encodes our naming conventions with `something-something-else-vDATESTAMP-somepostfix` and I am not in favor of this.

apozsuse: OK, I'll change the code to receive the regex from the job definition document vs creating it in mash.

smarlowucf: Shouldn't mash be told exactly what file and where to pull it from? I don't think we want mash searching for images files via regex.

apozsuse: I'll close  the PR and  create a new one with the new approach.

PR #882, opened 2024-01-12
Includes the implementation of the S3Bucket download function

apozsuse:
This PR implements the function called to download the image file to the download directory.

The function expects to have the complete url for the file to be downloaded in the `download_url` parameter. It will responsibility of the entity creating the mash job to provide the complete URL for the file to be downloaded for the image publication.

### DIFF
--- a/mash/services/download/s3bucket_job.py
+++ b/mash/services/download/s3bucket_job.py
@@ -26,6 +26,10 @@ from apscheduler.schedulers.background import BackgroundScheduler
 from apscheduler.events import EVENT_JOB_SUBMITTED
 # project
 from mash.services.base_defaults import Defaults
+from mash.utils.ec2 import (
+    get_session,
+    download_file_from_s3_bucket
+)
 
 
 class S3BucketDownloadJob(object):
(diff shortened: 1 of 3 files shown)


rjschwei (approved):

mash/services/download/s3bucket_job.py
```diff
@@ -180,3 +224,12 @@ def _job_skipped_event(self, event):
         # event was scheduled. In this case we just skip the event
         # and keep the active job waiting for an obs change
         pass
+
+    def _get_bucket_name_and_key_from_download_url(self) -> (str, str):
+        """Returns the bucket name and s3 object key from download_url param"""
```
smarlowucf, line 229: Nitpick: an example download_url could be helpful for reading the code.

mash/services/download/s3bucket_job.py
```diff
@@ -180,3 +224,12 @@ def _job_skipped_event(self, event):
...
+        s3_prefix = 's3://'
+        download_url = self.download_url
+        if download_url.startswith(s3_prefix):
+            download_url = download_url[len(s3_prefix):]
+        bucket_name, _, object_key = download_url.partition('/')
```
smarlowucf, line 234: Nitpick: Will this handle virtual directories? I.e. if the blob name uses forward slashes? I'm not familiar with the partition method. I think the use of forward slashes as delimiters is a frequently used technique. An example could be:

```
mybucket-123/amazon/image-123-20220202.tar.gz
```

Where a left split (max=1) on the slashes would provide the bucket name and the blob name. And a right split (max=1) would yield the tarball filename. E.g.

```
bucket=mybucket-123
blob=amazon/image-123-20220202.tar.gz
file=image-123-20220202.tar.gz
```

I think it's okay as it is. But if a blob is passed in that has delimiter slashes in the name it will cause an unhandled exception here during the tuple unpacking.

smarlowucf (approved):

mash/services/download/s3bucket_job.py
smarlowucf: NM 🤦🏼 . Wasn't familiar with the partition method. So the nitpick here would be the filename could end up with slashes which would cause problems during storage.

I.e. it try to store the example file in `/path/to/download/dir/amazon/image-123-20220202.tar.gz` which would cause file not found error.

apozsuse: Fixed the function just to get the last part of the file and keep the downloaded file path simple rather than having multiple subdirectories.

PR #883, opened 2024-01-17
Removes `last_service` overwrite in API for aws MP jobs

apozsuse:
`last_service` field was being overwritten to `publish` for aws MP jobs.

As we're sending the changelogs to AWS API for MP jobs in `deprecate` service, we need it not to be overwritten.


rjschwei (approved):

smarlowucf:
Good catch, thanks!

smarlowucf (approved):

PR #888, opened 2024-01-31
Remove the filename from download_url for S3Downloads

apozsuse:
In order to have a more similar behavior in both download types, for `S3DownloadJobs` this change removes the filename from the `download_url` param and the filename will be obtained from the `image` parameter in the mash job doc.

So, for the download of an image from `s3://download_bucket/directory1/directory2/image_file.tar.gz` we'd expect the following values in the mash job doc:
- `download_url`: `s3://download_bucket/directory1/directory2`
- `image`:  `image_file.tar.gz`

Tests have been modified accordingly

### DIFF
--- a/mash/services/download/s3bucket_job.py
+++ b/mash/services/download/s3bucket_job.py
@@ -198,7 +198,7 @@ class S3BucketDownloadJob(object):
                     destination_file
                 )
             )
-            self.image_filename = self.image_name
+            self.image_filename = destination_file
 
             # job finished successfully
             self.job_status = 'success'
--- a/mash/services/download/s3bucket_job.py
+++ b/mash/services/download/s3bucket_job.py
@@ -46,7 +46,7 @@ class S3BucketDownloadJob(object):
 
     * :attr:`download_url`
       S3 bucket URL.
-      Includes the `directory` part of the URL iif the key object contains some
+      Includes the `directory` part of the URL if the key object contains some
       directory structure.
 
     * :attr:`image_name`
--- a/mash/services/download/s3bucket_job.py
+++ b/mash/services/download/s3bucket_job.py
@@ -45,7 +45,13 @@ class S3BucketDownloadJob(object):
       job file containing the job description
 
     * :attr:`download_url`
-      S3 bucket URL
+      S3 bucket URL.
+      Includes the `directory` part of the URL iif the key object contains some
+      directory structure.
+
+    * :attr:`image_name`
+      Image name for the download.
+      Contains the filename for the image to be downloaded.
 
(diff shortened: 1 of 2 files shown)


mash/services/download/s3bucket_job.py
```diff
@@ -45,7 +45,13 @@ class S3BucketDownloadJob(object):
       job file containing the job description
 
     * :attr:`download_url`
-      S3 bucket URL
+      S3 bucket URL.
+      Includes the `directory` part of the URL iif the key object contains some
```
rjschwei, line 49: s/iif/if/

rjschwei (approved):

mash/services/download/s3bucket_job.py
```diff
@@ -163,27 +169,36 @@ def _download_image_file(self):
...
                 self._get_bucket_name_and_key_from_download_url()
 
             destination_file = os.path.join(
                 self.download_directory,
-                filename
+                self.image_name
             )
+
+            if dir_part_of_object_key:
+                if dir_part_of_object_key.endswith('/'):
```
smarlowucf, line 181: Nitpick/suggestion: Is the goal here to prevent concatenation causing `//`? If so this can be handled by using `os.path.normpath`.

mash/services/download/s3bucket_job.py
```diff
@@ -227,22 +242,21 @@ def _job_skipped_event(self, event):
...
-            /directory/my_file_name.tar.gz,
-            my_file_name.tar.gz
+            my-bucket-name ,
+            /directory1/directory2
         """
         s3_prefix = 's3://'
         download_url = self.download_url
         if download_url.startswith(s3_prefix):
             download_url = download_url[len(s3_prefix):]
         download_url_parts = download_url.split('/')
```
smarlowucf, line 256: Using `maxsplit=1` here would make line 260 easier to read as it then would be `download_url_parts[1]`.

mash/services/download/s3bucket_job.py
```diff
@@ -227,22 +242,21 @@ def _job_skipped_event(self, event):
...
+            s3://my-bucket-name/directory1/directory2
         It will return:
-            s3://my-bucket-name ,
-            /directory/my_file_name.tar.gz,
-            my_file_name.tar.gz
+            my-bucket-name ,
+            /directory1/directory2
         """
         s3_prefix = 's3://'
         download_url = self.download_url
```
smarlowucf, line 253: Nitpick: this could be a one liner: download_url = self.download_url.replace('s3://', '')

smarlowucf (approved):

PR #893, opened 2024-02-20
Changes `OBSDownloadJob` to be initialized over job_config and config

apozsuse:
Note that this PR includes the changes from [PR 892](https://github.com/SUSE-Enceladus/mash/pull/892)

In order to have the Download Service with a Factory pattern like the rest of the services, it is necessary to make some refactoring in the initialization of the 2 download classes. This PR refactors the initialization of the `OBSDownloadJob`.

### DIFF
--- a/mash/services/base_config.py
+++ b/mash/services/base_config.py
@@ -138,23 +138,13 @@ class BaseConfig(object):
 
         return data
 
-    def get_service_names(self, credentials_required=False):
+    def get_service_names(self):
         """
         Return a list of all service names.
-
-        If credentials_required is True return only services that require
-        credentials to execute.
+.
         """
         services = self._get_attribute(attribute='services') or \
             Defaults.get_service_names()
-
-        if credentials_required:
-            non_cred_services = self._get_attribute(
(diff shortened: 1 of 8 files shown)


mash/services/base_config.py
```diff
@@ -138,23 +138,13 @@ def get_cloud_data(self):
 
         return data
 
-    def get_service_names(self, credentials_required=False):
+    def get_service_names(self):
```
rjschwei, line 141: There is still an option that a service in the pipeline exists that does not need any credentials. From this change I cannot determine if treating all services the same at this level has any effect on the selection process. If we drop the `non_cred_services` config option, which I am not opposing, then this change would require a major version bump of mash when this gets released.

apozsuse: Closing this PR for now as it was not a good idea to mix both changes. I'll redo the PR just containing the OBSDownloadJob initialization change.

For the `non_cred_services` thing, we can discuss in [PR 892](https://github.com/SUSE-Enceladus/mash/pull/892)

PR #894, opened 2024-02-20
Changes `OBSDownloadJob` to be initialized over job_config and config

apozsuse:
In order to have the Download Service with a Factory pattern like the rest of the services, it is necessary to make some refactoring in the initialization of the 2 download classes. This PR refactors the initialization of the `OBSDownloadJob`.

### DIFF
--- a/mash/services/download/obs_job.py
+++ b/mash/services/download/obs_job.py
@@ -27,7 +27,8 @@ from apscheduler.events import EVENT_JOB_SUBMITTED
 from obs_img_utils.api import OBSImageUtil
 
 # project
-from mash.services.base_defaults import Defaults
+from mash.log.filter import BaseServiceFilter
+from mash.utils.mash_utils import setup_rabbitmq_log_handler
 
 
 class OBSDownloadJob(object):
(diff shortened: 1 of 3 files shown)


mash/services/download/obs_job.py
```diff
@@ -88,39 +89,77 @@ class OBSDownloadJob(object):
...
-        self.job_id = job_id
-        self.job_file = job_file
-        self.download_directory = os.path.join(download_directory, job_id)
-        self.download_url = download_url
-        self.image_name = image_name
-        self.last_service = last_service
+    def __init__(self, job_config, config):
+        self.job_config = job_config
+        self.config = config
+        self.job_id = job_config['id']
```
smarlowucf, line 95: For the required args we could wrap them in a try/except for keyerror. And raise a more helpful message about the required parameter. Otherwise these will fail on "KeyError: 'somearg'" which is hard to debug.

mash/services/download/obs_job.py
```diff
@@ -88,39 +89,77 @@ class OBSDownloadJob(object):
...
+            config.get_amqp_pass()
+        )
+        logger.addHandler(rabbit_handler)
+        logger.addFilter(BaseServiceFilter)
+        self.log_callback = logging.LoggerAdapter(
+            logger,
+            {'job_id': self.job_id}
+        )
+
+        if 'conditions' in job_config:
```
smarlowucf, line 120: Can these be simplified with "get's"? `self.conditions = job_config.get('conditions')`

smarlowucf (approved):

mash/services/download/obs_job.py
rjschwei: I agree all the conditionals should be dropped, let the dict object do the work
`self.arch = job_config.get('cloud_architecture', 'x86_64')` then we do not have to do the membership checks and don't need so many forks in the code.

rjschwei (changes requested):

rjschwei (approved):

smarlowucf (approved):

PR #895, opened 2024-02-20
Migrate to new version of azure-img-utils

smarlowucf:
This uses the new graph api which requires arg changes.


apozsuse (approved):
LGTM :+1:

rjschwei (approved):

PR #897, opened 2024-02-22
Flask sql alchemy paginate fix

schaefi:
This change is two fold

**Fixing type error of paginate**
    
As of Flask-SQLAlchemy >= 3.0 all arguments to paginate are keyword-only. This commit changes the call to be keyword based and thus continues to work on older and newer Flask-SQLAlchemy

**Update tox.ini to work in this century**

For running the tests and actually using tox in newer versions, these changes are required due to tox being more restrictive in newer versions

### DIFF
--- a/mash/services/database/utils/jobs.py
+++ b/mash/services/database/utils/jobs.py
@@ -63,8 +63,8 @@ def get_jobs(user_id, page=1, per_page=10):
     Retrieve all jobs for user.
     """
     job_query = Job.query.filter_by(user_id=user_id).paginate(
-        page,
-        per_page,
+        page=page,
+        per_page=per_page,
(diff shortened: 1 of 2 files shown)


schaefi:
For reference I'm running mash in a container using SUSE's new code base, from which these changes originates from. For details see: https://build.opensuse.org/project/show/Virtualization:Appliances:Images

I'm also working on a quick start documentation update PR since mash works nicely now from a container instance and I think this can help to lift mash also on the side of more up-to-date distros.

smarlowucf (approved):
LGTM

rjschwei (approved):

smarlowucf:
Thanks @schaefi

PR #898, opened 2024-02-22
Update documentation

schaefi:
Added a Quick Start chapter which describes how to use the mash OCI container. I reworked the OBS project that builds the mash container to use the latest code base and to be actually useful again. Project can be found here:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images
 
This makes using mash and hacking on mash an easier start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,27 @@ From here the following setup steps are required:
        "utctime": "now"
    }
    EOF
+   ```
 
+   __NOTE__:
+    The setting for boot_firmware connects to the boot capabilities of the
+    referenced image and should be taken seriously. An image to boot in the
+    AWS cloud can be selected to use either a BIOS+MasterBootRecord interface
+    or an EFI+EFI-binary interface. Images built by SUSE supports both modes
+    and we prefer EFI boot over BIOS boot which is the reason for the above
+    setting. In case a self-built image or other image source should be used
+    with mash, it's important to clarify on the boot capabilities first.
+
+   To put the job into the pipeline call the following command:
+
+   ```bash
    mash job ec2 add example.job
+   ```
+
+   List your job(s) with the following command:
+
+   ```bash
+   mash job list
    ```
 
 ## Installation
--- a/README.md
+++ b/README.md
@@ -7,6 +7,89 @@ It is written in Python3 and designed to run as a set of Systemd & HTTP
 services. Mash currently supports a number of Public Cloud Frameworks; such
 as Amazon EC2, Google Compute Engine and Microsoft Azure.
 
+## Quick Start
+
+For a quick start into mash and its service pipeline, an OCI container
+image is provided. The container can be used as a mash development
+environment as well as a server instance for integration testing of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,89 @@ It is written in Python3 and designed to run as a set of Systemd & HTTP
...
+   ```bash
+   cat >example.job <<-EOF
+   {
+       "boot_firmware": [
+           "uefi-preferred"
+       ],
+       "cloud_account": "private",
+       "cloud_architecture": "x86_64",
+       "cloud_image_name": "EOS-Update-Server-alp-x86_64-v{date}",
+       "download_url": "https://ftp.gwdg.de/pub/opensuse/repositories/home:/marcus.schaefer:/EOS/images_ALP/",
```
schaefi, line 79: I'm not happy with the example referenced here. It's one of my images that I used for testing but for the purpose of the mash docu  I think it would be nice to use a different one. I just could not came across a good example. Do you have an idea ?

schaefi:
I also noticed the following. Only projects that defines

```
PublishFlags: withreports
```

can be used by mash because it requires to either find ```.packages``` or ```.report```. This also means no other URL source than an OBS project repository can be consumed by mash. Reading the code I think it would be possible to let it work without these metadata files in place.

I think at least in the docs we should mention that. If wished I can add an information here ?

smarlowucf (approved):
Thanks, @schaefi we could use an openSUSE image as the example.

smarlowucf:
> es in place.
> 
> I think at least in the docs we should mention that. If wished I can add an information here ?

There are two options now. Images can be pulled from s3 buckets or obs download pages. The specifics of pulling from a download page are embedded in [obs-img-utils](https://github.com/SUSE-Enceladus/obs-img-utils). We should be able to make a change there that removes the requirement on the metadata files existing. However, if any image package conditions are used then metadata is required.

README.md
rjschwei: Also it might be good to make a comment about `uefi-preferred`. For the example we do not really want to give big explanations of he values, but in this case it is probably warranted.

rjschwei (approved):

README.md
schaefi: Thanks and done. I added a reference to the opensuse image rather than my personal one, thanks @smarlowucf 
@rjschwei I added a note about the importance of the boot firmware interface and its connection to the image binary. 

Feel free to change my wording if need be :)

schaefi:
> > es in place.
> > I think at least in the docs we should mention that. If wished I can add an information here ?
> 
> There are two options now. Images can be pulled from s3 buckets or obs download pages. The specifics of pulling from a download page are embedded in [obs-img-utils](https://github.com/SUSE-Enceladus/obs-img-utils). We should be able to make a change there that removes the requirement on the metadata files existing. However, if any image package conditions are used then metadata is required.

I added a PR to the obs-img-utils to make that happen and added you as a reviewer.

rjschwei:
LGTM, thanks

PR #900, opened 2024-02-23
Fix parameter name with the image name (it's `image` not `image_name`)

apozsuse:


smarlowucf:
@apozsuse your fix looks good. I added another patch if you can please review that change.

smarlowucf (approved):

rjschwei (approved):

smarlowucf:
@apozsuse merging to get fixes deployed but please still give it a once over when you have a chance.

PR #902, opened 2024-02-28
Create blob url for image blob with compute cred

smarlowucf:
If there is a partner center cred provided it will not have permission to perform compute/storage actions.


apozsuse (approved):
LGTM :+1:

rjschwei (commented):
LGTM

PR #904, opened 2024-03-01
Adds support to imds_version parameter in Mash for ec2 images

apozsuse:
Additionally updates the requirements over the version of `ec2imgutils` supporting this feature.

### DIFF
--- a/mash/services/create/ec2_job.py
+++ b/mash/services/create/ec2_job.py
@@ -73,6 +73,7 @@ class EC2CreateJob(MashJob):
             'boot_firmware',
             ['uefi-preferred']
         )
+        self.imds_version = self.job_config.get('imds_version', '')
 
         # EC2 images only support one firmware
         self.boot_firmware = self.boot_firmware[0]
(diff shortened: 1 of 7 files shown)


smarlowucf (approved):
LGTM, but this cannot be merged until mash is package for python 3.11 as well.

rjschwei (approved):

apozsuse:
Updated the PR to the current main branch.

mash/services/create/ec2_job.py
```diff
@@ -73,6 +73,7 @@ def post_init(self):
             'boot_firmware',
             ['uefi-preferred']
         )
+        self.imds_version = self.job_config.get('imds_version', '')
```
smarlowucf, line 76: We could set the default here to 'v2.0' which simplifies most jobs that would have to set that value manually.

mash/services/jobcreator/ec2_job.py
```diff
@@ -51,6 +51,7 @@ def post_init(self):
             False
         )
         self.launch_inst_type = self.kwargs.get('launch_inst_type')
+        self.imds_version = self.kwargs.get('imds_version', '')
```
smarlowucf, line 54: Ditto

smarlowucf (approved):

smarlowucf:
Good to merge once conflict is resolved.

smarlowucf (approved):

PR #905, opened 2024-03-04
Replacing `setup` functions for `setup_method` in tests classes as `setup` seems deprecated by pytest

apozsuse:
Apparently [setup is now deprecated](https://docs.pytest.org/en/latest/deprecations.html#setup-teardown)


smarlowucf (approved):
Thanks!

PR #906, opened 2024-03-27
Add validation for sku and generation_id

smarlowucf:
If these are the same fail early. This prevents image creation service running into a confusing error.


rjschwei (approved):

apozsuse (approved):
LGTM :+1:

PR #911, opened 2024-05-14
Modifies the ec2 accounts API to include the `subnets` element

apozsuse:
As discussed in the previous PR, in order to maintain the multi-tenancy feature of mash we need multiple `subnet` elements for every account to be able to perform tests in different regions.

Thus, this change modifies the API for the `ec2` accounts API endpoint to support such feature.
The format for the parameter is the following:

```
...
'subnets': [
       {
             'region': 'us-east-99',
             'subnet': 'subnet-000099'
        },
        {
             'region': 'us-east-100',
             'subnet': 'subnet-000100'
        }
```

For now this PR just addresses the API change. The DB  changes will be implemented in the following PR.
`mash-client` will be modified accordingly to this change (if required)

### DIFF
--- a/mash/services/api/v1/schema/accounts/ec2.py
+++ b/mash/services/api/v1/schema/accounts/ec2.py
@@ -40,7 +40,7 @@ partition = {
     'enum': ['aws', 'aws-cn', 'aws-us-gov']
 }
 
-ec2_subnets = {
+test_regions = {
     'type': 'array',
     'items': {
         'type': 'object',
@@ -63,7 +63,7 @@ ec2_account = {
         'group': string_with_example('group1'),
         'partition': partition,
         'region': string_with_example('us-east-1'),
-        'subnets': ec2_subnets
+        'test_regions': test_regions
     },
     'additionalProperties': False,
     'required': [
--- a/mash/services/api/v1/schema/accounts/ec2.py
+++ b/mash/services/api/v1/schema/accounts/ec2.py
@@ -40,6 +40,21 @@ partition = {
     'enum': ['aws', 'aws-cn', 'aws-us-gov']
 }
 
+ec2_subnets = {
+    'type': 'array',
+    'items': {
+        'type': 'object',
+        'properties': {
+            'region': string_with_example('us-east-44'),
+            'subnet': string_with_example('T-12345678')
+        },
+        'required': ['region', 'subnet'],
+        'additionalProperties': False
+    },
+    'minItems': 1,
+    'example': [{'region': 'us-east-44', 'subnet': 'subnet-12345678'}]
+}
+
 ec2_account = {
     'type': 'object',
     'properties': {
@@ -48,7 +63,7 @@ ec2_account = {
         'group': string_with_example('group1'),
         'partition': partition,
         'region': string_with_example('us-east-1'),
-        'subnet': string_with_example('subnet-12345678')
+        'subnets': ec2_subnets
     },
     'additionalProperties': False,
     'required': [
@@ -79,7 +94,7 @@ ec2_account_update = {
         'additional_regions': additional_regions,
         'group': string_with_example('group1'),
         'region': string_with_example('us-east-1'),
-        'subnet': string_with_example('subnet-12345678'),
+        'subnets': ec2_subnets,
         'credentials': ec2_credentials
     },
     'additionalProperties': False,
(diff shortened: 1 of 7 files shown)


mash/services/api/v1/utils/jobs/ec2.py
```diff
@@ -64,8 +64,9 @@ def add_target_ec2_account(
     """
     job_doc_data = cloud_accounts.get(account['name'], {})
     region_name = job_doc_data.get('region') or account.get('region')
-    subnet = job_doc_data.get('subnet') or account.get('subnet')
-
+    subnet = job_doc_data.get('subnet') or get_subnet_for_region(
```
rjschwei, line 67: I think keeping `subnet` in the jobdoc and adding `subnets` creates a condition where we need to some cross-reference checking.  As implemented here they appear to be mutually exclusive, which means we would need to produce an error or warning somewhere that is `subnet` is provided the information in `subnets` is ignored.

Also we are creating the situation where one setting is implicit, which is the current implementation, i.e. it is assumed that the provided `subnet` exists in `us-east-1`, and for `subnets` we are being explicit. I think we should consider dropping `subnet` from the jobdoc and only supporting `subnets`

smarlowucf: The existing account model has both region option and subnet. The assumption is that the subnet provided exists in the region provided. The new model would make regions a mapping so any number of regions/subnets could be provided. Anything coming from the job doc would be treated as the override. But that could get complicated with a mapping.

mash/services/api/v1/schema/accounts/ec2.py
```diff
@@ -48,7 +63,7 @@
         'group': string_with_example('group1'),
         'partition': partition,
         'region': string_with_example('us-east-1'),
-        'subnet': string_with_example('subnet-12345678')
+        'subnets': ec2_subnets
```
smarlowucf, line 66: I'd suggest this be named regions. That way if there's any other regional attribute we want in the future we could use the same mapping. subnet in this context is kind of an attr of the region.

mash/services/api/v1/schema/accounts/ec2.py
```diff
@@ -40,6 +40,21 @@
     'enum': ['aws', 'aws-cn', 'aws-us-gov']
 }
 
+ec2_subnets = {
```
smarlowucf, line 43: Might be helpful for this to be regions and a dict where the keys are region names. Then each region is a sub dict with subnet. That way it's easy to get a list of region names if it's needed.

mash/services/api/v1/utils/jobs/ec2.py
apozsuse: OK, I'll think about how to integrate  all the `regions` related fields into a single entity and come with a proposal.

mash/services/api/v1/schema/accounts/ec2.py
apozsuse: I checked the `flask_restx` models support to have such structure and it seems [maintainers don't recommend this approach](https://github.com/python-restx/flask-restx/issues/524) but a list of objects with fixed fields for better marshalling and validation. I'll stick to the list of fixed objects  for now...

smarlowucf: That's right I had trouble with that before. Sounds good. 👍🏼

mash/services/api/v1/utils/jobs/ec2.py
apozsuse: Update the PR with the following changes:
- As the `region` and `subnet` elements are used for the image creation, after some thought, I think it's better leave them as they are, so each account will have a `region` and a `subnet` that will be used to create the image.
- additionally a new entity `test_regions` has been added. This new entity contains the additional regions and subnets that will be used for the tests if present.
- This `test_regions` is included in the `target_account_info` param for the jobs

mash/services/api/v1/schema/accounts/ec2.py
apozsuse: As there's some `additional_regions` element and I think it's better not to modify the `region` and `subnet` elements, called the entity `test_regions`, as I think it's more descriptive about the intended usage for that information.

mash/services/api/v1/schema/accounts/ec2.py
apozsuse: Finally used the list of objects for the `test_regions` element.

smarlowucf (approved):

PR #912, opened 2024-05-17
Database modification to have new ec2_subnet table

apozsuse:
In order to maintain the multitenancy of mash we need to store the different `subnet` for different regions for the same `ec2_account`.

A modification in the database model is implemented to remove the `subnet` column from the `ec2_account` table and include a new `ec2_subnet` table that contains the `region`, `subnet` and `account_id`.

This way a single `ec2_account` can store multiple `subnet`s for different `region`s.

The apropriate db migration script is included in the PR.


rjschwei (approved):

smarlowucf:
Is this ready for review and match changes from #911 ?

apozsuse:
> Is this ready for review and match changes from #911 ?

Nope, not yet. I'll push new  changes  and ping  you for review

apozsuse:
Updated the PR with the new approach to the `ec2_test_region`  table.

One doubt I had was if it might be good to populate, in the migration script, a first ec2_test_region if the account has `region` and `subnet` already populated. I did not include it, as the account will need to set up additional test regions, but it could be a good feature for simple use cases where the account just wants to test in one region...

If you think it'd be useful, let me know and I'll include it in the migration script.

smarlowucf:
This is mainly to add the data field? There will be a separate PR for usage in test service?

smarlowucf (approved):

apozsuse:
> This is mainly to add the data field? There will be a separate PR for usage in test service?

Yep, this one was just the database changes, there will be more PRs to address all the changes (I'm trying to keep the PRs as small as possible).

PR #914, opened 2024-05-31
Use explicit requirements on 3.11

smarlowucf:
To preserve package name


smarlowucf:
Requirements don't resolve properly using python macros

rjschwei (approved):

PR #919, opened 2024-07-16
Update Quick Start documentation

schaefi:
Further changes on the mash-client allows to pre-setup the mash user and login via a systemd service in the container. This simplifies running a job in the containerized startup of mash. This commit updates the Quick Start documentation regarding the changes on the mash container side.


schaefi:
For reference: https://github.com/SUSE-Enceladus/mash-client/pull/94

smarlowucf (approved):

PR #921, opened 2024-08-21
Adds the test_preparation service based on replicate service class

apozsuse:

This PR is part of the improvement effort in the test strategy. This PR implements the test_preparation service, which is the service that is executed right before the test service. The main goal for this test_preparation service is to copy the created image into the test_regions configured in the database for the account, so some tests (with different instance types/configurations) can be executed in these additional regions.

This implementation reuses the job class for the replicate service, as the functionality is the same.

The tests have been tested in a local development account against my aws personal account.

Nope

### Additional Information

This new service is really very similar to the replicate service.
Unit tests have been included for the service.


mash/services/replicate/ec2_job.py
```diff
@@ -38,12 +38,23 @@ def post_init(self):
...
-                self.job_config['replicate_source_regions']
+            # Boolean that indicates if we're using the class in the replicate
+            # service (False, default value) or in the test_preparation service
+            # (True)
+            self.test_preparation = \
+                self.job_config.get('test_preparation', False)
+            if self.test_preparation:
+                # Class used in the test_preparation service
+                self.replicate_source_regions = \
+                    self.job_config['test_preparation_regions']
```
smarlowucf, line 49: I wonder if we could simplify this more by making the variable name more generic and use it for both replicate and test prep? Thus minimizing the need for conditionals.

mash/services/test_preparation_service.py
```diff
@@ -22,9 +22,11 @@
 
 # project
 from mash.mash_exceptions import MashException
+from mash.services.no_op_job import NoOpJob
 from mash.services.test_preparation.config import TestPreparationConfig
 from mash.services.listener_service import ListenerService
 from mash.services.job_factory import BaseJobFactory
+from mash.services.test_preparation.ec2_job import EC2TestPreparationJob
```
smarlowucf, line 29: Should this be the replicate job class?

mash/services/replicate/ec2_job.py
apozsuse: The only problem I see not to use the exact `replicate` class is to keep track  of the `original` region where the image was created in, so the `test_cleanup` service cleans everything but that image (for the replicate service to copy the image after the tests are completed).

Let me check if the original region can be identified somehow...

apozsuse: Modified the PR to change as little as possible the EC2ReplicateJob class.
 Now the struct of the message that triggers the test_preparation job is the same as the one used for the replicate job, so the only difference is the key used to store the result in the `status_msg` dictionary.
I'd  want to keep that data separated (the ami-ids for the images replicated and which  regions) to make it easier for the `test_cleanup` service identify which images  are to be removed after the tests.

smarlowucf (approved):
LGTM

PR #922, opened 2024-08-27
Adds the code for the test_cleanup service for EC2 jobs

apozsuse:

This PR includes the code for the new `test_cleanup` service. This service is responsible to remove the images that have been replicated in the configured test regions after tests are complete.

The job config document for the job has to include a `test_cleanup_regions` entry to get the account corresponding to the different new regions the images where replicated to by the `test_preparation` service.

Additionally gets the ami ids for the new images in those regions from the status_msg as published there by the `test_preparation` service.

The tests have been locally tested with a mash development instance.

### DIFF
--- a/mash/services/test_cleanup/ec2_job.py
+++ b/mash/services/test_cleanup/ec2_job.py
@@ -1,0 +1,72 @@
+# Copyright (c) 2019 SUSE LLC.  All rights reserved.
+#
+# This file is part of mash.
+#
+# mash is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# mash is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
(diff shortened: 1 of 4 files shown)


mash/services/test_cleanup/ec2_job.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2019 SUSE LLC.  All rights reserved.
```
smarlowucf, line 1: 2024

smarlowucf (approved):
For catalog jobs that will replicate the image to all regions does this end up causing duplication? E.g. if we copy into a couple test regions then clean the images up then replicate back to the same regions. If so not sure if it's that big of a deal to worry about but just a thought.

apozsuse:
> For catalog jobs that will replicate the image to all regions does this end up causing duplication? E.g. if we copy into a couple test regions then clean the images up then replicate back to the same regions. If so not sure if it's that big of a deal to worry about but just a thought.

I've been doing some tests creating a Leap image in my AWS account and running multiple tests the same day to the same regions did not give me any error about duplications.

The cleanup code wipes out the image and the snapshot, and a new image with  the same name can be created by the replicate service. I'll have a look if we need to implement some waiting code to assure the image is actually deleted before finishing the `test_cleanup` service.

I' doing real tests with my account, so if there's any problem with the replication after the cleanup we'll detect it in these tests.

PR #930, opened 2024-12-05
Implements the instance selector for the test service

apozsuse:

Implements the instance selection from the catalog to cover the feature_combinations that are required to be tested.

### DIFF
--- a/mash/services/test/ec2_test_utils.py
+++ b/mash/services/test/ec2_test_utils.py
@@ -107,11 +107,6 @@ def select_instances_for_tests(
     """
     instances = []
     for feature_combination in feature_combinations:
-        if logger:
-            logger.debug(
-                'Selecting instance for feature combination: '
-                f'{feature_combination}'
-            )
         instance = select_instance_for_feature_combination(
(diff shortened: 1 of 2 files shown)


mash/services/test/ec2_test_utils.py
```diff
@@ -86,3 +90,90 @@ def remove_incompatible_feature_combinations(feature_combinations):
...
+            )
+        instance = select_instance_for_feature_combination(
+            feature_combination=feature_combination,
+            instance_catalog=instance_catalog,
+            logger=logger
+        )
+        if instance:
+            instances.append(instance)
+            if logger:
+                logger.debug(f'The following instance was selected {instance}')
```
rjschwei, line 123: We can incorporate the feature combination used in this message and drop the earlier message. The selection process is fast as such I don't think we need an "announcement message" in the log

smarlowucf (approved):

mash/services/test/ec2_test_utils.py
apozsuse: Fixed! Thanks!

PR #931, opened 2024-12-12
Add file package to spec

smarlowucf:
This is required for processing the files on disk and is not always available in container builds


apozsuse (approved):
LGTM

PR #933, opened 2024-12-20
Include region/partition criteria for test instance selection

apozsuse:
This PR includes the partition/region as a criteria for the selection of the instances that will be used for the tests.

The mechanism will try to test all the instance feature combinations possible that can be covered for each partition with the test instance catalog configured. If no test can be executed in some partition  (with exceptions as arm64 images in `aws-cn` or `aws-us-gov` partitions) an exception is raised (as there's some config issue in the instance catalog).

I still left the actual test execution commented. I'll test it and include it in the following PR.

Also updated the provided configuration example with actual values of instances for each region.

### DIFF
--- a/mash/services/test/ec2_job.py
+++ b/mash/services/test/ec2_job.py
@@ -30,7 +30,8 @@ from mash.services.test.utils import (
 )
 from mash.services.test.ec2_test_utils import (
     get_instance_feature_combinations,
-    select_instances_for_tests
+    select_instances_for_tests,
+    get_partition_test_regions
 )
 from mash.utils.mash_utils import create_ssh_key_pair
 from mash.utils.ec2 import (
@@ -84,6 +85,13 @@ class EC2TestJob(MashJob):
             )
             return
 
+        self.partitions = get_partition_test_regions(self.test_regions)
+        if not self.partitions:
+            msg = 'At least one partition is required for tests.'
+            if self.log_callback:
+                self.log_callback.error(msg)
+            raise MashTestException(msg)
+
         self.cloud_architecture = self.job_config.get(
             'cloud_architecture', 'x86_64'
         )
@@ -94,25 +102,6 @@ class EC2TestJob(MashJob):
         self.cpu_options = self.job_config.get('cpu_options', {})
         self.instance_catalog = self.config.get_test_ec2_instance_catalog()
 
-        self.feature_combinations = get_instance_feature_combinations(
-            self.cloud_architecture,
-            self.boot_firmware,
-            self.cpu_options,
-            logger=self.log_callback
-        )
-        self.instance_types = select_instances_for_tests(
-            instance_catalog=self.instance_catalog,
-            feature_combinations=self.feature_combinations,
-            logger=self.log_callback
-        )
-
-        if not self.instance_types:
-            msg = (
-                'No instances in the instance catalog can cover the required '
-                f'feature combinations: {self.feature_combinations}'
-            )
-            raise MashTestException(msg)
-
         # Get all account credentials in one request
         accounts = []
         for region, info in self.test_regions.items():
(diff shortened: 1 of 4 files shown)


mash/services/test/ec2_job.py
```diff
@@ -94,58 +102,72 @@ def run_job(self):
...
-            instance_catalog=self.instance_catalog,
-            feature_combinations=self.feature_combinations,
-            logger=self.log_callback
-        )
 
-        if not self.instance_types:
-            msg = (
-                'No instances in the instance catalog can cover the required '
-                f'feature combinations: {self.feature_combinations}'
+        # mash will try to test all the possible features supported in the
```
smarlowucf, line 121: Does this allow for a job to only publish to one partition? In that case we wouldn't want to test the other partitions.

mash/services/test/ec2_job.py
```diff
@@ -94,58 +102,72 @@ def run_job(self):
...
+            if not instance_types:
 
-        for instance_type in self.instance_types:
-            if self.log_callback:
-                self.log_callback.info(
-                    'Running img-proof tests against image with '
-                    'type: {inst_type}.'.format(
-                        inst_type=instance_type
-                    )
+                if partition in ('aws-cn', 'aws-us-gov') and \
```
smarlowucf, line 134: Could this clause be moved up one level? There's no need to check for types if this condition exists.

mash/services/test/ec2_job.py
```diff
@@ -94,58 +102,72 @@ def run_job(self):
...
-            logger=self.log_callback
-        )
 
-        if not self.instance_types:
-            msg = (
-                'No instances in the instance catalog can cover the required '
-                f'feature combinations: {self.feature_combinations}'
+        # mash will try to test all the possible features supported in the
+        # available test regions for each partition
+        for partition, test_regions in self.partitions:
```
smarlowucf, line 123: I think we need to do some smoke testing with cn partition in particular. If this will start to cause a lot of unnecessary false positives. The networking in CN region can be troublesome which leads to false job failures.

mash/services/test/ec2_job.py
```diff
@@ -84,6 +85,13 @@ def run_job(self):
             )
             return
 
+        self.partitions = get_partition_test_regions(self.test_regions)
+        if not self.partitions:
+            msg = 'At least one partition is required for tests.'
```
rjschwei, line 90: This is really a configuration issue, i.e. the `partition` key missing in the config. The message should express that this needs to be fixed in the configuration.

mash/services/test/ec2_job.py
```diff
@@ -94,58 +102,72 @@ def run_job(self):
...
-                self.log_callback.info(
-                    'Running img-proof tests against image with '
-                    'type: {inst_type}.'.format(
-                        inst_type=instance_type
-                    )
+                if partition in ('aws-cn', 'aws-us-gov') and \
+                        self.cloud_architecture == 'aarch64':
+                    # Skip test aarch64 images in China and GovCloud.
+                    # There are no aarch64 based instance types available.
+                    continue
```
rjschwei, line 138: This is no longer correct, gov-cloud has `M6g` and `T4g` [1] we should fix this, meaning push aarch64 images to the gov-cloud regions. `T4g` is also available in China [2]

[1] https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-instance-regions.html
[2] https://www.amazonaws.cn/en/ec2/instance-types/

mash/services/test/ec2_job.py
```diff
@@ -94,58 +102,72 @@ def run_job(self):
...
+
+                # There are no instance types configured in the test regions
+                # for the partition that can cover a single feat combination
+                msg = (
+                    f'No instances in the instance catalog for {partition}'
+                    f' partition can cover the any feature combination: '
```
rjschwei, line 144: s/any//

rjschwei (changes requested):

mash/services/test/ec2_job.py
apozsuse: According to [this comment](https://github.com/SUSE-Enceladus/mash/pull/933#discussion_r1901244238) from Robert, this is no longer the case and we should remove it.
I'll update this PR removing this (and updating the instance catalog in the proposed configuration).
Then I'll check what has to be changed to start pushing aarch64 images to Cn and sovereign regions.

mash/services/test/ec2_job.py
apozsuse: Not sure I follow your comment.
With the code as in this PR, all the instance  features are tested in every partition, and that could be a problem, as you say, in CN region if tests fail per the unreliable behavior of the partition itself. Furthermore, a exception is raised if some feature combination is NOT possible to be tested in some partition.

Giving it a second thought, I was wondering if it would be better, to avoid too much job false failures, to try to test the features in a `best effort` basis (removing [this exception raise](https://github.com/SUSE-Enceladus/mash/blob/341c563aba552efb211c4a14c9c4d0fe21651058/mash/services/test/ec2_test_utils.py#L131) and just writing a warning log if some feature cannot be tested in some partition) for us to  control what is tested or not by means of the test instance catalog config.  
That way we'd control, for example,  if SEV is tested in CN or not, just removing the feature as supported in the CN test instance catalog... and we'd just a warning in the log stating SEV cannot be tested in CN partition with the current config...

wdyt?

mash/services/test/ec2_job.py
apozsuse: Do we have an easy way to test a `T4g` instance with some image in the region?

mash/services/test/ec2_job.py
apozsuse: Fixed.

mash/services/test/ec2_job.py
apozsuse: Fixed.

mash/services/test/ec2_job.py
smarlowucf: I did a quick check with the cli:

```
14:40:58 ▶ aws ec2 describe-instance-type-offerings --region cn-north-1 --profile {profile_name} | grep t4g
            "InstanceType": "t4g.micro",
            "InstanceType": "t4g.small",
            "InstanceType": "t4g.nano",
            "InstanceType": "t4g.xlarge",
            "InstanceType": "t4g.medium",
            "InstanceType": "t4g.large",
            "InstanceType": "t4g.2xlarge",
```

mash/services/test/ec2_job.py
smarlowucf: The trouble with only treating failures as warning level is that something could slip through the cracks. Technically the partitions are different reference implementations and thus there's a chance an image fails on one and no on another. Probably the best we can do is to note this concern when we deploy. That way if we notice an uptick in false failures we are aware of the reason.

rjschwei (approved):

smarlowucf (approved):

PR #936, opened 2025-05-02
Updates the instance test types in Azure

apozsuse:
Updating the instance types used to test the images in Azure to more recent ones.


smarlowucf (approved):
Thanks!

PR #938, opened 2025-05-07
Migrate to gceimutils

smarlowucf:

Migrate gce utilities to gceimgutils and drop dependency on gce rest api client.

### Additional Information

This relies on a new version of gceimgtuils being released with all the recent pr's


apozsuse (approved):
LGTM

apozsuse:
Always good to have PRs removing way more lines that the ones added! Thanks Sean!

PR #945, opened 2025-05-26
Includes the cpu_options param in the img-proof invocation

apozsuse:

This PR includes the required cpu_options parameter when creating the test instance that will run the tests. This allows us to include the testing of instance features like the confidential compute ones.

This change has been tested locally with a mash instance running against my aws account.


apozsuse:
Merging this PR to update main branch in this feature branch.

PR #951, opened 2025-08-06
Implements the function to select the different instance configuration for tests

apozsuse:

This PR includes the function that selects, for a group of feature instance combinations, the instance configurations that will be used to run the tests over the image that will be published.

This change will be tested locally with a personal gcp account.

### DIFF
--- a/mash/services/test/gce_test_utils.py
+++ b/mash/services/test/gce_test_utils.py
@@ -244,7 +244,7 @@ def select_instance_config_for_feature_combination(
         selected_group = random.choice(candidate_groups)
         instance_config = {
             'arch': arch,
-            'instance_name': random.choice(selected_group['instance_names']),
+            'instance_type': random.choice(selected_group['instance_types']),
             'boot_type': boot_type,
             'region': selected_group['region'],
             'shielded_vm': shielded_vm,
--- a/mash/services/test/gce_test_utils.py
+++ b/mash/services/test/gce_test_utils.py
@@ -153,7 +153,7 @@ def remove_incompatible_feature_combinations(feature_combinations):
     return feature_combinations
 
 
-def select_instances_for_tests(
+def select_instance_configs_for_tests(
     instance_catalog: list,
     feature_combinations: list,
     logger: logging.Logger = None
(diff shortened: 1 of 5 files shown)


mash/services/test/gce_test_utils.py
```diff
@@ -145,3 +151,103 @@ def remove_incompatible_feature_combinations(feature_combinations):
     for incompatible_combination in incompatible_combinations:
         feature_combinations.remove(incompatible_combination)
     return feature_combinations
+
+
+def select_instances_for_tests(
```
smarlowucf, line 156: This could be a little confusing an instance vs an instance type. I think in these functions the thing that is in focus would be instance type not instance?

smarlowucf: If so I suggest including that in naming. E.g. select_instance_types_for_tests. etc

mash/services/test/gce_test_utils.py
```diff
@@ -145,3 +151,103 @@ def remove_incompatible_feature_combinations(feature_combinations):
...
+        instance = select_instance_for_feature_combination(
+            feature_combination=feature_combination,
+            instance_catalog=instance_catalog,
+            logger=logger
+        )
+        if instance:
+            instances.append(instance)
+            if logger:
+                logger.debug(
+                    f'Selected instance {instance} for {feature_combination}'
```
smarlowucf, line 180: Ditto anywhere the intent is to reference an "instance type"

mash/services/test/gce_test_utils.py
```diff
@@ -145,3 +151,103 @@ def remove_incompatible_feature_combinations(feature_combinations):
...
+        ):
+            # required confidential_compute feature not supported
+            continue
+
+        candidate_groups.append(instance_group)
+
+    instance = None
+    if candidate_groups:
+        selected_group = random.choice(candidate_groups)
+        instance = {
```
smarlowucf, line 244: Maybe something like s/instance/instance_config would be more explicit

mash/services/test/gce_test_utils.py
```diff
@@ -145,3 +151,103 @@ def remove_incompatible_feature_combinations(feature_combinations):
...
+            'arch': arch,
+            'instance_name': random.choice(selected_group['instance_names']),
```
smarlowucf, line 246: s/instance_name/instance_type?

smarlowucf:
I think referencing instance types as "instances" can cause confusion. Consider using "instance type" terminology instead.

mash/services/test/gce_test_utils.py
apozsuse: Changed it to instance_config. I think it's the most accurate name. Thx!

mash/services/test/gce_test_utils.py
apozsuse: Done!

Only changed it for GCE files. I'll change it for aws in another PR.
Thx, Sean!

PR #952, opened 2025-08-06
Improves instance selection variables to improve code legibility

apozsuse:

This PR changes the name of some variables and functions for EC2 tests. The goal is to have better self describing variables and fns as we are using those to get an `instance_config`  (the configuration for the instance for some specific test) rather than just the instance type.

Tests will be executed in a local mash instance with personal cloud accounts.


smarlowucf (approved):
LGTM 👍🏼 Thanks!

Issue #872, opened 2023-12-18
Handle AWS Marketplace Jobs With multiple accounts


smarlowucf:
Currently the AWS MP path allows for an account list but this is not handled properly. An attempt to pass multiple accounts would lead to unexpected failure. The relevant job classes should be updated to properly handle multiple accounts.

PR #978, opened 2026-04-03
Handle a mapping of catalog to entity id

smarlowucf:
This allows for a configurable catalog per entity id. It allows allows for updating multiple entities in the same account with the same image.

### DIFF
--- a/mash/services/api/v1/utils/jobs/ec2.py
+++ b/mash/services/api/v1/utils/jobs/ec2.py
@@ -117,7 +117,6 @@ def convert_account_dict(accounts):
 
 def validate_mp_fields(job_doc):
     mp_fields = [
-        'entity_id',
         'version_title',
         'release_notes',
         'access_role_arn',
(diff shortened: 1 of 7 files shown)


smarlowucf:
Note: there is an existing bug that mash does not handle an account list properly for MP jobs https://github.com/SUSE-Enceladus/mash/issues/872. Given there is a plan to pull out the marketplace handling I chose to make these changes under the same assumption that the accounts list only ever contains one account.

mash/services/api/v1/utils/jobs/ec2.py
```diff
@@ -140,6 +139,18 @@ def validate_mp_fields(job_doc):
...
+            'One of entity_id or entity_ids is required to publish a '
+            'marketplace image and both are missing in the job doc.'
+        )
+
+    if job_doc.get('entity_id') and job_doc.get('entity_ids'):
+        raise MashJobException(
+            'Only one of entity_id or entity_ids can be provided to publish '
+            'a marketplace image and both are in the job doc.'
+        )
+
```
rjschwei, line 153: Keeping `entity_id` for backwards compatibility for now makes sense. That said I think we should have a deprecation warning when `entity_id` is used and give a time when that option is going to be removed. In the future, whenever we decide that date will be, maybe 1 year from now, we would then only accept `entity_ids` after all a list of one is still only 1 item.

The deprecation notice text should then also be included in the exception messages.

mash/services/deprecate/ec2_mp_job.py
```diff
@@ -81,28 +95,35 @@ def run_job(self):
...
+                    catalog=entity['catalog']
                 )
-            )
+
+                change_set_id = response.get('ChangeSetId')
+                self.status_msg['change_set_ids'][entity_id] = change_set_id
+                self.log_callback.info(
+                    'Marketplace change set submitted. Change set id: '
+                    '{change_set}'.format(
+                        change_set=change_set_id
```
rjschwei, line 125: WE should probably add the catalog name to the message to make tracking easier.

rjschwei (approved):

PR #980, opened 2026-04-24
Fix bug in image tags functionality

smarlowucf:
Mash can accept the image tags as a dictionary object which allows for better data validation.


apozsuse (approved):
LGTM. Thanks, Sean!

PR #983, opened 2026-05-21
Drop version pin on flask

smarlowucf:
flask-restx now supports new versions of flask


apozsuse:
If I remember properly, I think there were breaking changes with Flask 3.x, are those gone?

smarlowucf:
Yeah, it was flask-restx that was causing trouble. It's fixed now to work with Flask >= 3.0.0

apozsuse (approved):
Thx!

PR #984, opened 2026-05-27
Add labels option to GCE job creation

smarlowucf:
Allow specifying a dictionary of key/value label pairs that are applied to the image when creating it in GCE. The labels are passed through from the API schema to the jobcreator and on to GCECreateImage in the create service.


apozsuse (approved):
LGTM. Thx!

rjschwei (approved):

